PR #163, opened 2018-02-16
update NOTICE.txt dependencies list

sfalquier:


NOTICE.txt
```diff
@@ -1,2 +1,30 @@
+
 ARLAS-server
-Copyright 2017 Gisaïa
\ No newline at end of file
+Copyright 2017 and onwards Gisaïa
```
MohamedHamouGisaia: 2017 - 2018 ?

sfalquier: `2017 and onwards` requires less maintenance each year ;)

PR #165, opened 2018-02-27
fix doc generation using swagger.json instead of swagger.yml

sfalquier:
Quick fix proposal.


sfalquier:
The less inelegant solution I found today to fix this problem both on CI and local runs.
Let me know what you think about it.

sylvaingaudan:
LGTM

PR #195, opened 2018-03-13
Add WFS suite test to CI

### LINKED ISSUE #172: Plug OGC wfs test suite docker in ARLAS Server CI

mbarbet:
fix #172

### DIFF
--- a/scripts/tests-integration.sh
+++ b/scripts/tests-integration.sh
@@ -87,3 +87,36 @@ docker run --rm \
 	--net arlas_default \
 	maven:3.5.0-jdk-8 \
 	mvn install -DskipTests=false
+
+echo "===> run integration WFS tests"
+docker run --rm \
+    -w /opt/maven \
+    -v $PWD:/opt/maven \
+    -v $HOME/.m2:/root/.m2 \
+    -e ARLAS_HOST="arlas-server" \
+    -e ARLAS_PORT="9999" \
+    -e ARLAS_PREFIX="/arlas/" \
+    -e ARLAS_ELASTIC_HOST="elasticsearch" \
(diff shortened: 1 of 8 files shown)


src/test/java/io/arlas/server/WFSTool.java
```diff
@@ -0,0 +1,9 @@
+package io.arlas.server;
+
+public class WFSTool {
```
sfalquier: This class is not direcly linked to WFS stuff.
It should be renamed like `CollectionTool`. This class should wrap code from `AbstractTestWithCollection.beforeClass` and `AbstractTestWithCollection.afterClass`. This code would be called in main method according to args[] values (triggered from maven with something like `-Dexec.args="load"` or `-Dexec.args="delete"`).
`AbstractTestWithCollection` would call `CollectionTool` methodes in `beforeClass` and `afterClass`.

scripts/tests-integration.sh
```diff
@@ -87,3 +87,23 @@ docker run --rm \
...
+    -v $PWD:/opt/maven \
+    -v $HOME/.m2:/root/.m2 \
+    -e ARLAS_HOST="arlas-server" \
+    -e ARLAS_PORT="9999" \
+    -e ARLAS_PREFIX="/arlas/" \
+    -e ARLAS_ELASTIC_HOST="elasticsearch" \
+    -e ARLAS_ELASTIC_PORT="9300" \
+    --net arlas_default \
+    maven:3.5.0-jdk-8 \
+    mvn exec:java -Dexec.mainClass="io.arlas.server.WFSTool" -Dexec.classpathScope=test
```
sfalquier: You may clean collection and .arlas indeces after tests in case new tests are appended right after WFS ones.

PR #196, opened 2018-03-14
fix#180 improve WFSException

### LINKED ISSUE #180: Add Throwable cause to the constructor of WFSException

mbarbet:
fix #180


src/main/java/io/arlas/server/exceptions/WFSException.java
```diff
@@ -41,7 +65,7 @@ public WFSException(WFSExceptionCode exceptionCode, String exceptionText, String
     }
 
     public WFSException(WFSExceptionCode exceptionCode, String exceptionText) {
-        this(exceptionCode, exceptionText, null);
+        this(exceptionCode, exceptionText, "null");
```
sfalquier: Nothing more relevant than "null" for locator value when not provided?

PR #201, opened 2018-03-16
Make application context path configurable

### LINKED ISSUE #158: Configuration: applicationContextPath can be customized in a container via an environment variable
in `ARLAS-server/conf/configuration.yaml`, the variable `applicationContextPath` can be customized in a container via an environment variable

sfalquier:
fixes #158


sfalquier:
Tests widely modified to take into account app context path.

PR #202, opened 2018-03-16
fix #194

### LINKED ISSUE #194: Avoid "." in xml element name of a WFS response
Replace "." in xml element name of a WFS response by a configurable character

mbarbet:
fix #194

### DIFF
--- a/src/main/java/io/arlas/server/wfs/utils/XmlUtils.java
+++ b/src/main/java/io/arlas/server/wfs/utils/XmlUtils.java
@@ -31,6 +31,9 @@ import java.util.Stack;
 
 public class XmlUtils {
 
+
+    public static String pointPathSubstitute;
+
     public static final String ELEMENT = "element";
     public static final String NAME = "name";
     public static final String TYPE = "type";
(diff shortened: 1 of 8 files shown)


src/main/java/io/arlas/server/wfs/utils/XmlUtils.java
```diff
@@ -98,8 +117,13 @@ private static void writeElementForType(XMLStreamWriter writer, String nameToDis
     }
 
     private static void writeElement(XMLStreamWriter xmlStream, String nameToDisplay, String value,String uri,String prefix) throws XMLStreamException {
+        nameToDisplay=cleanName(nameToDisplay);
         xmlStream.writeStartElement(prefix, nameToDisplay, uri);
         xmlStream.writeCharacters(value);
         xmlStream.writeEndElement();
     }
+
```
sfalquier: You should remove these lines.

src/main/java/io/arlas/server/wfs/utils/XmlUtils.java
```diff
@@ -31,6 +31,9 @@
 
 public class XmlUtils {
 
+
+    public static String charToUse;
```
sfalquier: Could you find a more explicit name for this var? Maybe something mentioning "separator" or something like that. On top of that, related methods cleanName and retrieveOriginalName could also be more explicit.

PR #205, opened 2018-03-20
generate arlas-api using typescript-fetch

### LINKED ISSUE #199: Generate arlas-api with typescript-fetch

MohamedHamouGisaia:
fix #199 
L'image `gisaia/swagger-codegen:2.3.1` est déjà poussé dans DockerHub

### DIFF
--- a/Dockerfile-swagger-codegen
+++ b/Dockerfile-swagger-codegen
@@ -5,5 +5,5 @@ ADD swagger-codegen/modules/swagger-codegen-cli/target/swagger-codegen-cli.jar /
 
 CMD java -jar /opt/swagger/swagger-codegen-cli.jar generate \
    -i /opt/gen/target/tmp/swagger.json \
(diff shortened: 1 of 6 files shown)


MohamedHamouGisaia:
@mbarbet @sebbousquet @sfalquier @sylvaingaudan , should we change the name of arlas-api ?

sfalquier:
Are you aware of the fact that bumping swagger codegen to 2.3.1 will change drastically generated API and we have to change ARLAS web contributors/components?

MohamedHamouGisaia:
Yes. The generated API keeps the same names of classes and the same attributes. The only difference for typescript-fetch is that Http responses are Promises instead of Observables. For this, few changes were made in ARLAS-web-core to convert promises to observables. 
The api generated for typescript-fetch using swagger codegen 2.2.3 was different from the angular one

PR #235, opened 2018-04-06
Format code and organize imports

sfalquier:


sfalquier:
Intelli-J conf : 
Settings → Editor → General → Auto Import → Java → Optimize imports on the fly
Code → Reformat code

PR #241, opened 2018-04-09
Feature/implement get records

mbarbet:


src/main/java/io/arlas/server/dao/ElasticCollectionReferenceDaoImpl.java
```diff
@@ -132,6 +132,37 @@ public CollectionReference getCollectionReference(String ref) throws ArlasExcept
         }
     }
 
+    @Override
+    public List<CollectionReference> getCollectionReferences(String[] includes, String[] excludes, int size, int from, String filter) throws ArlasException {
```
sfalquier: It looks like filter is not used and should be removed, right?

src/main/java/io/arlas/server/ogc/csw/ElasticCSWService.java
```diff
@@ -0,0 +1,4 @@
+package io.arlas.server.ogc.csw;
+
+public class ElasticCSWService {
```
sfalquier: This class doesn't seem to be directly linked to this commit.

src/main/java/io/arlas/server/ogc/csw/utils/ElementSetName.java
```diff
@@ -0,0 +1,4 @@
+package io.arlas.server.ogc.csw.utils;
+
+public enum ElementSetName {
```
sfalquier: This class doesn't seem to be directly linked to this commit.

sfalquier: even if it is modified in the next commit.

mbarbet: Fixed

PR #253, opened 2018-04-13
Set ARLAS_ELASTIC_NODES to empty string by default

sfalquier:


sfalquier:
@sylvain-gaudan could you have a look to the smallest PR I have ever made? ;o)

PR #258, opened 2018-04-16
Hot fix/fix check and dc

mbarbet:


src/main/java/io/arlas/server/dao/ElasticCollectionReferenceDaoImpl.java
```diff
@@ -266,7 +266,7 @@ private void checkCollectionReferenceParameters(CollectionReference collectionRe
     }
 
     private void checkExcludeField(CollectionReferenceParameters params, List<String> fields) throws NotAllowedException {
-        if (params.excludeFields != null) {
+        if (params.excludeFields != null&&params.excludeFields!="") {
```
sfalquier: Some white spaces are missing, aren't they?

PR #259, opened 2018-04-16
Support elasticsearch aliases as collection's index name

sfalquier:


src/test/java/io/arlas/server/ogc/wfs/WFSServiceIT.java
```diff
@@ -92,14 +95,19 @@ public void handleHeaderFilter(ValidatableResponse then) throws Exception {
...
 
     public void handleDescribeFeature(ValidatableResponse then) throws Exception {
-        then.statusCode(200)
-                .body("xs:schema.complexType.complexContent.extension.sequence.element.size()", equalTo(7));
+        if(!DataSetTool.ALIASED_COLLECTION) {
+            then.statusCode(200)
+                    .body("xs:schema.complexType.complexContent.extension.sequence.element.size()", equalTo(7));
+        } else {
+            then.statusCode(200)
+                    .body("xs:schema.complexType.complexContent.extension.sequence.element.size()", equalTo(9));
```
sfalquier: @mbarbet Could you have a closer look on this? I don't understand why it may be 7... and why it turns to 9 with only  one additional field in the mapping. I am not sure of what these xml elements represent.

mbarbet:
FFTM

PR #264, opened 2018-04-24
Helm packaging

elouanKeryell-Even:


packaging/helm/arlas-server/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+appVersion: 5.6.0
```
sfalquier: Is it possible to get this value automatically or to insert it through `release.sh` process like we have already done for API_VERSION ?
If we keep it hardcoded, we will have to update this file before each release manually... and probably forget to do it ;)

packaging/helm/README.md
```diff
@@ -0,0 +1,11 @@
+[Helm](helm.sh) is a package management system for kubernetes applications.
+
+Once you have Helm installed, you can install the arlas-server chart by performing the following:
+
+```bash
+helm install -f my_values.yaml arlas-server
+```
+
+, where `my_values.yaml` contains custom configuration settings for the chart.
```
sfalquier: Perhaps you should also add a new section in docs/arlas-server-configuration.md about this conf file?

packaging/helm/arlas-server/LICENSE.txt
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
sfalquier: Should be : 
`Copyright 2017 and onwards Gisaïa`

packaging/helm/arlas-server/Chart.yaml
sfalquier: Do you think You coudl handle this before tomorrow evening? I would like to release arlas-server on friday and it would be nice to have helm packaging included.

elouanKeryell-Even: should be fixed by 1af5c55bf439f7e44c5cbf095a20485c03a74bc5 (see [here](https://github.com/gisaia/ARLAS-server/blob/1af5c55bf439f7e44c5cbf095a20485c03a74bc5/release.sh#L125-L126)).

packaging/helm/README.md
elouanKeryell-Even: Done in 1af5c55bf439f7e44c5cbf095a20485c03a74bc5, see [here](https://github.com/gisaia/ARLAS-server/blob/1af5c55bf439f7e44c5cbf095a20485c03a74bc5/docs/arlas-server-configuration.md#helm-configuration).

packaging/helm/arlas-server/LICENSE.txt
elouanKeryell-Even: Fixed in 1af5c55bf439f7e44c5cbf095a20485c03a74bc5, see [here](https://github.com/gisaia/ARLAS-server/blob/1af5c55bf439f7e44c5cbf095a20485c03a74bc5/packaging/helm/arlas-server/LICENSE.txt#L190).

packaging/helm/arlas-server/Chart.yaml
sfalquier: Ok, but you have to replace FULL_API_VERSION by ARLAS_VERSION for these files (basically, next ARLAS_VERSION=5.6.1 while FULL_API_VERSION=5.0.1) and you should remove hardcoded reference to `5.6.0` in your yml files to avoid confusion.
By the way, we will release helm packaging in 5.6.2. I would be safer ;)

elouanKeryell-Even: Should be fixed in ec502d3797f114decd0e8d89ef1d2bb0f946e724

elouanKeryell-Even:
Squashed commits.

PR #278, opened 2018-04-30
Implement range service 

### LINKED ISSUE #269: Implement a `_range` service that calculates the min and max values of a given field

MohamedHamouGisaia:
This service calculates the min and max values of a given field and given filters.
fixes #269


src/main/java/io/arlas/server/core/FluidSearch.java
```diff
@@ -102,6 +104,9 @@
     public static final String RANGE_ALIASES_CHARACTER = "$";
     public static final String TIMESTAMP_ALIAS = "timestamp";
 
+    public static final String MIN_VALUE_OF_FILED = "min_field_values";
```
sfalquier: Should be `field_min_value`, shouldn't it?

MohamedHamouGisaia: Yes, it's better

PR #281, opened 2018-05-14
Split maven project into a multi modules project

sfalquier:


sfalquier:
@mbarbet Could you have a look to this job failure?
https://travis-ci.org/gisaia/ARLAS-server/jobs/378565742

ogc-csw/src/main/java/io/arlas/server/ogc/csw/CSWService.java
```diff
@@ -69,7 +70,9 @@
 import static io.arlas.server.utils.CheckParams.isBboxLatLonInCorrectRanges;
 
 
-public class CSWService extends CollectionRESTServices {
+public class CSWService {
+
+    public Logger LOGGER = LoggerFactory.getLogger(CSWService.class);
```
MohamedHamouGisaia: LOGGER is not used

MohamedHamouGisaia: Sorry, it's actually used

mbarbet: it's a forgotten, we must remove these uses, it was for debugging

PR #288, opened 2018-05-17
Feat/#287 python api generation

TamerGisaia:
Fix #288


scripts/build-swagger-codegen.sh
```diff
@@ -1,6 +1,5 @@
-#!/bin/bash
-
-SWAGGER_VERSION="2.3.1"
+SWAGGER_VERSION="2.2.3"
+API_LANGUAGE="python"
```
sfalquier: Since `python` is hardcoded, we won't be able to generate a new version of the docker image for typescript.
`API_LANGUAGE` (and maybe `SWAGGER_VERSION`) shoud be passed as build-args (https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg)

Dockerfile-swagger-codegen-python
```diff
@@ -0,0 +1,11 @@
+FROM openjdk:8-jre-alpine
```
sfalquier: Instead of having multipe Docker-swagger-codegen-* files, it would be better to have a generic one that accept `LANGUAGE` as a build argument (https://docs.docker.com/engine/reference/commandline/build/#set-build-time-variables---build-arg).

scripts/build-swagger-codegen.sh
```diff
@@ -1,6 +1,5 @@
-#!/bin/bash
-
-SWAGGER_VERSION="2.3.1"
+SWAGGER_VERSION="2.2.3"
```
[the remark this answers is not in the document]
TamerGisaia: 2.3.1 generate invalid python api that's why we need to use 2.2.3. 
This problem might be fixed in future released of Swagger-Codegen.

Dockerfile-swagger-codegen-python
TamerGisaia: The problem here is that we have different commands for each of the docker files. For python docker file we need to have "python-config.json" and we need to use it with option '-c'.

As far as I know, making this with build arguments is not possible. Instead I made changes "build-swagger-codegen.sh" that makes it accept two arguments "v for version and l for language".

sfalquier: Ok, let's keep it like that.

sfalquier:
Feel free to merge.

PR #298, opened 2018-06-06
Include url_template of opensearch in ARLAS server configuration

### LINKED ISSUE #286: OPENSEARCH: add url_template_prefix in ARLAS server configuration instread of collection reference definition
OPENSEARCH: Remove `url_template_prefix` from collection reference definition and specifiy it in the configuration of ARLAS server instead. 
The `url_template_prefix` must contain a placemark denoted as `{collection}` so that ARLAS Server can inject the collection name.
Also, documentation in `arlas-server-configuration.md` should describe the role of `{collection}` in `url_template_prefix`

MohamedHamouGisaia:
Fixes #286

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/app/OpensearchConfiguration.java
+++ b/arlas-core/src/main/java/io/arlas/server/app/OpensearchConfiguration.java
@@ -17,26 +17,14 @@
  * under the License.
  */
 
-package io.arlas.server.rest.explore.opensearch.model;
+package io.arlas.server.app;
 
-import javax.xml.bind.annotation.XmlAttribute;
+import com.fasterxml.jackson.annotation.JsonProperty;
 
-public class Url {
+public class OpensearchConfiguration {
 
-    public static enum REL {results, suggestions, self, collection}
+    public static final String COLLECTION_PLACEMARK = "COLLECTION";
 
-    @XmlAttribute(name = "template")
-    public String template = "";
-
-    @XmlAttribute(name = "type")
-    public String type = "";
-
-    @XmlAttribute(name = "rel")
-    public REL rel;
-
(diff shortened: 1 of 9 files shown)


arlas-core/src/main/java/io/arlas/server/app/OpensearchConfiguration.java
```diff
@@ -0,0 +1,30 @@
...
+ * under the License.
+ */
+
+package io.arlas.server.app;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+public class OpensearchConfiguration {
+
+    public static final String COLLECTION_PALCEMARK = "[collection]";
```
sfalquier: You'd better use `{` instead of `[`, it's more or less a standard...

conf/configuration.yaml
```diff
@@ -118,3 +118,9 @@ arlas-ogc:
 arlas-csw:
   openSearchDescription: ${ARLAS_CSW_OPENSEARCH_DESCRIPTION:-Geo-BigData Collection Catalog}
   openSearchShortName: ${ARLAS_CSW_OPENSEARCH_SHORTNAME:-Geo-Catalog}
+
+########################################################
+############ OPENSEARCH                  ###############
+########################################################
+opensearch:
+  url-template-prefix: ${ARLAS_OPENSEARCH_URL_TEMPLATE:-http://ARLAS-HOST/arlas/explore/[collection]/_search} # Only ARLAS_HOST and the endpoint '_search' can be modified. '/arlas/explore/[collection]/' should remain as it is.
```
sfalquier: I guess there is a missunderstanding about this point.
Devops that uses ARLAS should be able to use rewrite rules to expose their own url schemes. As a result, the whole `url-template-prefix` can be modified.

PR #340, opened 2018-08-29
Fix error logging on mappings union operation

### LINKED ISSUE #338: Describe collection generates inconsistant warn logs
Many logs are generated on many API calls. They look like :
`WARN : Cannot union fields [key FIELD_NAME] for collection describe`

Reason : when trying to union mappings of collection's indeces in `ElasticAdmin`, we generate a log for all fields that appears in all mappings.
https://github.com/gisaia/ARLAS-server/blob/develop/arlas-core/src/main/java/io/arlas/server/core/ElasticAdmin.java#L87

Impacted requests : 
  - REST (/arlas/explore/{collection}/_describe)
  - Opensearch (/ogc/opensearch/{collection})
  - WFS (/ogc/wfs/{collection})

TODO : fix `if` conditions so that we only warn for fields that are not consistant between indeces (object and value for the same FIELD_NAME in different indeces)

sfalquier:
fix #338

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/core/ElasticAdmin.java
+++ b/arlas-core/src/main/java/io/arlas/server/core/ElasticAdmin.java
@@ -81,10 +81,10 @@ public class ElasticAdmin {
         for (String key : update.keySet()) {
             if(!ret.containsKey(key)) {
                 ret.put(key,update.get(key));
+            } else if(ret.get(key).type != update.get(key).type) {
+                LOGGER.error("Cannot union field [key=" + key + "] because type mismatch between indices' mappings");
             } else if(ret.get(key).properties instanceof Map && update.get(key).properties instanceof Map) {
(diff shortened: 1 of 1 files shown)


arlas-core/src/main/java/io/arlas/server/core/ElasticAdmin.java
```diff
@@ -81,10 +81,10 @@ public CollectionReferenceDescription describeCollection(CollectionReference col
         for (String key : update.keySet()) {
             if(!ret.containsKey(key)) {
                 ret.put(key,update.get(key));
+            } else if(ret.get(key).type != update.get(key).type) {
+                LOGGER.error("Cannot union field [key=" + key + "] because type mismatch between indeces' mappings");
```
sylvaingaudan: indeces=>indices

PR #373, opened 2018-12-04
Initialize InspireConfiguration when it is not initialized in yaml conf

### LINKED ISSUE #376: Initialize InspireConfiguration when it is not initialized in yaml conf

MohamedHamouGisaia:
Fix #376

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/app/ArlasServerConfiguration.java
+++ b/arlas-core/src/main/java/io/arlas/server/app/ArlasServerConfiguration.java
@@ -31,6 +31,8 @@ import org.elasticsearch.common.Strings;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.time.LocalDate;
+import java.time.format.DateTimeFormatter;
 import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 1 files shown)


arlas-core/src/main/java/io/arlas/server/app/ArlasServerConfiguration.java
```diff
@@ -170,6 +170,13 @@ public void check() throws ArlasConfigurationException {
         if (arlasServiceExploreEnabled == null) {
             arlasServiceExploreEnabled = true;
         }
+        if (inspireConfiguration == null) {
+            inspireConfiguration = new InspireConfiguration();
+            inspireConfiguration.enabled = false;
+            inspireConfiguration.publicAccessLimitations = "no limitations apply";
+            inspireConfiguration.accessAndUseConditions = "no conditions apply";
+            inspireConfiguration.servicesDateOfCreation = "2018-11-27";
```
sfalquier: Strange to have a hardcoded date here.
Maybe create a date from `LocalDate.now()`?

MohamedHamouGisaia: It corresponds to the date we released v9.6.0 that supports Inspire compliant services

PR #374, opened 2018-12-04
Enhance Inspire parameters validity check & throw propre exceptions on collections import when Inspire config is enabled

MohamedHamouGisaia:


MohamedHamouGisaia:
MR re-edited and commits squashed :+1:

PR #399, opened 2019-02-12
Add dateformat parameter to explore api endpoints

### LINKED ISSUE #389: Add `format` option in [range, lt, gt, lte, gte] queries for date fields

MohamedHamouGisaia:
Fix #389

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/utils/CheckParams.java
+++ b/arlas-core/src/main/java/io/arlas/server/utils/CheckParams.java
@@ -30,6 +30,7 @@ import io.arlas.server.model.Keyword;
 import io.arlas.server.model.enumerations.*;
 import io.arlas.server.model.request.*;
 import io.arlas.server.model.response.RangeResponse;
+import org.joda.time.format.DateTimeFormat;
 
 import java.text.DateFormat;
 import java.text.ParseException;
@@ -37,7 +38,6 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
-import java.util.Locale;
 import java.util.regex.Pattern;
 
 public class CheckParams {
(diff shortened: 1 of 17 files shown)


arlas-core/src/main/java/io/arlas/server/utils/CheckParams.java
```diff
@@ -136,6 +136,17 @@ public static void checkFilter(Filter filter) throws ArlasException {
...
         }
+        checkDateFormat(filter.dateformat);
+    }
+
+    public static void checkDateFormat(String dateFormat) throws ArlasException {
+        if (!StringUtil.isNullOrEmpty(dateFormat)) {
+            try {
+                DateTimeFormat.forPattern(dateFormat);
+            } catch (IllegalArgumentException e) {
+                throw new InvalidParameterException("Invalid date format '" + dateFormat + "'");
```
sylvaingaudan: Add getMessage from the exception so that we know what is wrong with that format.

arlas-core/src/main/java/io/arlas/server/utils/ParamsParser.java
```diff
@@ -238,9 +240,24 @@ public static Filter getFilter(List<String> filters, List<String> q, List<String
...
         return filter;
     }
 
+    public static String parseDate(String dateValue, String dateFormat) throws ArlasException {
+        String parsedDate = dateValue;
+        if (!StringUtil.isNullOrEmpty(dateFormat)) {
+            try {
+                parsedDate = String.valueOf(DateTimeFormat.forPattern(dateFormat).withZoneUTC().parseDateTime(dateValue).getMillis());
+            } catch (DateTimeParseException e) {
+                throw new InvalidParameterException(dateValue + " doesn't match the date format '" + dateFormat + "'");
```
sylvaingaudan: Add getMessage from the exception so that we know what is wrong with that date & format.

arlas-core/src/main/java/io/arlas/server/utils/ParamsParser.java
```diff
@@ -238,9 +240,24 @@ public static Filter getFilter(List<String> filters, List<String> q, List<String
...
+            } catch (IllegalArgumentException e) {
+                throw new InvalidParameterException(dateValue + " doesn't match the date format '" + dateFormat + "'");
```
sylvaingaudan: Add getMessage from the exception so that we know what is wrong with that date & format.

MohamedHamouGisaia:
@sfalquier @mbarbet @sylvaingaudan 
I first added a method that checks if the field's type is `date` when the `dateformat` is set; then i dropped it. 
This way, If a timestamp is indexed as long for example, we still can query it even by giving the dateformat.
The query will work for all fields types actually as long as the given field value respects the `dateformat`.
Do you think we should add a restriction to the field's type ? Like accepting only fields which type are dates and long for example ?

arlas-core/src/main/java/io/arlas/server/utils/ParamsParser.java
MohamedHamouGisaia: Done :+1:

arlas-core/src/main/java/io/arlas/server/utils/ParamsParser.java
MohamedHamouGisaia: Done :+1:

arlas-core/src/main/java/io/arlas/server/utils/CheckParams.java
MohamedHamouGisaia: Done :+1:

PR #401, opened 2019-02-13
Add fetchHits parameter to aggregations

### LINKED ISSUE #391: Add a the possibility to fetch an attribute (first, last, default, ...) in aggregation responses

MohamedHamouGisaia:
Fix #391

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/core/FluidSearch.java
+++ b/arlas-core/src/main/java/io/arlas/server/core/FluidSearch.java
@@ -19,6 +19,12 @@
 
 package io.arlas.server.core;
 
+
+import io.arlas.server.utils.ElasticTool;
+import org.elasticsearch.search.aggregations.metrics.tophits.TopHitsAggregationBuilder;
+import org.locationtech.jts.geom.*;
+import org.locationtech.jts.io.ParseException;
+import org.locationtech.jts.io.WKTReader;
 import io.arlas.server.exceptions.*;
 import io.arlas.server.model.CollectionReference;
 import io.arlas.server.model.enumerations.*;
@@ -28,7 +34,6 @@ import io.arlas.server.model.request.Metric;
 import io.arlas.server.model.request.MultiValueFilter;
 import io.arlas.server.model.response.TimestampType;
 import io.arlas.server.utils.CheckParams;
-import io.arlas.server.utils.ElasticTool;
 import io.arlas.server.utils.ParamsParser;
 import io.arlas.server.utils.StringUtil;
 import org.elasticsearch.action.search.SearchRequestBuilder;
@@ -50,7 +55,6 @@ import org.elasticsearch.search.aggregations.bucket.terms.IncludeExclude;
 import org.elasticsearch.search.aggregations.bucket.terms.TermsAggregationBuilder;
 import org.elasticsearch.search.aggregations.metrics.max.MaxAggregationBuilder;
 import org.elasticsearch.search.aggregations.metrics.min.MinAggregationBuilder;
-import org.elasticsearch.search.aggregations.metrics.tophits.TopHitsAggregationBuilder;
 import org.elasticsearch.search.aggregations.support.ValuesSourceAggregationBuilder;
 import org.elasticsearch.search.sort.SortBuilders;
 import org.elasticsearch.search.sort.SortOrder;
@@ -59,6 +63,7 @@ import org.locationtech.jts.io.ParseException;
 import org.locationtech.jts.io.WKTReader;
 import org.locationtech.jts.operation.valid.IsValidOp;
 import org.locationtech.jts.operation.valid.TopologyValidationError;
+
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -88,6 +93,7 @@ public class FluidSearch {
     public static final String HISTOGRAM_AGG = "Histogram aggregation";
     public static final String TERM_AGG = "Term aggregation";
     public static final String GEOHASH_AGG = "Geohash aggregation";
+    public static final String FETCH_HITS_AGG = "fetched_hits";
     public static final String GEOHASH_AGG_WITH_GEOASH_STRATEGY = "Geohash aggregation with geoash strategy";
     public static final String GEO_DISTANCE = "geodistance";
     public static final String NOT_ALLOWED_AS_MAIN_AGGREGATION_TYPE = " aggregation type is not allowed as main aggregation. Please make sure that geohash or term is the main aggregation or use '_aggregate' service instead.";
@@ -595,6 +601,7 @@ public class FluidSearch {
         dateHistogramAggregationBuilder = dateHistogramAggregationBuilder.dateHistogramInterval(intervalUnit);
         //get the field, format, collect_field, collect_fct, order, on
         dateHistogramAggregationBuilder = (DateHistogramAggregationBuilder) setAggregationParameters(aggregationModel, dateHistogramAggregationBuilder);
+        dateHistogramAggregationBuilder = (DateHistogramAggregationBuilder) setHitsToFetch(aggregationModel, dateHistogramAggregationBuilder);
         return dateHistogramAggregationBuilder;
     }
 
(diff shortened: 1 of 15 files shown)


arlas-core/src/main/java/io/arlas/server/core/FluidSearch.java
```diff
@@ -729,6 +731,34 @@ private ValuesSourceAggregationBuilder setAggeragatedGeometryStrategy(Aggregatio
         return aggregationBuilder;
     }
 
+    private ValuesSourceAggregationBuilder setHitsToFetch(Aggregation aggregationModel, ValuesSourceAggregationBuilder aggregationBuilder) throws ArlasException {
+        if (aggregationModel.fetchHits != null) {
+            TopHitsAggregationBuilder topHitsAggregationBuilder = AggregationBuilders.topHits("fetched_hits");
+            Integer size = aggregationModel.fetchHits.size != null ? aggregationModel.fetchHits.size : 1;
```
laurent-thiebaud-gisaia: You may use an Optional to avoid the null check

arlas-core/src/main/java/io/arlas/server/core/FluidSearch.java
```diff
@@ -729,6 +731,34 @@ private ValuesSourceAggregationBuilder setAggeragatedGeometryStrategy(Aggregatio
...
+                        topHitsAggregationBuilder.sort(field.substring(1), SortOrder.ASC);
+                    } else if (field.startsWith("-")) {
+                        includes.add(field.substring(1));
+                        topHitsAggregationBuilder.sort(field.substring(1), SortOrder.DESC);
+                    } else {
+                        includes.add(field);
+                    }
+                });
+                String[] hitsToInclude = new String[includes.size()];
+                hitsToInclude = includes.toArray(hitsToInclude);
```
laurent-thiebaud-gisaia: You can probably instanciate the array in one line, like includes.toArray(new String[includes.size()])

arlas-core/src/main/java/io/arlas/server/services/ExploreServices.java
```diff
@@ -257,8 +254,13 @@ public AggregationResponse formatAggregationResult(MultiBucketsAggregation aggre
                     if (subAggregationResponse.name.equals(FluidSearch.TERM_AGG)) {
                         subAggregationResponse.sumotherdoccounts = ((Terms) subAggregation).getSumOfOtherDocCounts();
                     }
-
-                    if (subAggregation.getName().equals(FluidSearch.DATEHISTOGRAM_AGG) || subAggregation.getName().startsWith(FluidSearch.GEOHASH_AGG) || subAggregation.getName().equals(FluidSearch.HISTOGRAM_AGG) || subAggregation.getName().equals(FluidSearch.TERM_AGG)) {
+                    if (subAggregation.getName().equals("fetched_hits")) {
```
laurent-thiebaud-gisaia: This seems like a bad use of map, you're adding something to a list from a forEach, can this not be replaced by a map?

arlas-core/src/main/java/io/arlas/server/core/FluidSearch.java
```diff
@@ -754,6 +758,31 @@ private ValuesSourceAggregationBuilder setAggeragatedGeometryStrategy(Aggregatio
...
     }
 
+    private ValuesSourceAggregationBuilder setHitsToFetch(Aggregation aggregationModel, ValuesSourceAggregationBuilder aggregationBuilder) throws ArlasException {
+        if (aggregationModel.fetchHits != null) {
+            TopHitsAggregationBuilder topHitsAggregationBuilder = AggregationBuilders.topHits(FETCH_HITS_AGG);
+            Integer size = Optional.ofNullable(aggregationModel.fetchHits.size).orElse(1);
+            topHitsAggregationBuilder.size(size);
+            List<String> includes = new ArrayList<>();
+            if (aggregationModel.fetchHits.include != null) {
+                for (String field : aggregationModel.fetchHits.include) {
```
laurent-thiebaud-gisaia: You muse a .map instead of a list that you feed from a for() loop

arlas-core/src/main/java/io/arlas/server/dao/ElasticCollectionReferenceDaoImpl.java
```diff
@@ -153,59 +153,24 @@ public void deleteCollectionReference(String ref) throws NotFoundException, Inte
...
+        List<String> fields = new ArrayList<>();
+        if (collectionReference.params.idPath != null)
+            fields.add(collectionReference.params.idPath);
+        if (collectionReference.params.geometryPath != null)
+            fields.add(collectionReference.params.geometryPath);
+        if (collectionReference.params.centroidPath != null)
+            fields.add(collectionReference.params.centroidPath);
+        if (collectionReference.params.timestampPath != null)
+            fields.add(collectionReference.params.timestampPath);
+        if(collectionReference.params.excludeFields!=null && collectionReference.params.excludeFields!=""){
```
laurent-thiebaud-gisaia: If you have Apache, there is a StringUtils.isEmpty or .isBlank to check a string for not null nor empty

arlas-core/src/main/java/io/arlas/server/dao/ElasticCollectionReferenceDaoImpl.java
```diff
@@ -153,59 +153,24 @@ public void deleteCollectionReference(String ref) throws NotFoundException, Inte
...
+            List<String> excludeField = Arrays.asList(collectionReference.params.excludeFields.split(","));
+            CheckParams.checkExcludeField(excludeField, fields);
+        }
+        ElasticTool.checkAliasMappingFields(client, collectionReference.params.indexName, collectionReference.params.typeName, fields.toArray(new String[fields.size()]));
+        List<String> indeces = ElasticTool.getIndecesName(client, collectionReference.params.indexName, collectionReference.params.typeName);
```
laurent-thiebaud-gisaia: is "indeces" the correct typing?

arlas-core/src/main/java/io/arlas/server/utils/ElasticTool.java
```diff
@@ -77,6 +80,48 @@ public static boolean checkIndexMappingFields(Client client, String index, Strin
         return true;
     }
 
+    public static boolean checkAliasMappingFields(Client client, String alias, String typeName, String... fields) throws ArlasException {
+        List<String> indeces = ElasticTool.getIndecesName(client, alias, typeName);
+        for (String index : indeces) {
```
laurent-thiebaud-gisaia: You may write it in one line: indeces.forEach(index => checkIndexMappingFields(client, index, typeName, fields);)

arlas-core/src/main/java/io/arlas/server/utils/ParamsParser.java
```diff
@@ -105,6 +111,21 @@ public static Aggregation getAggregationModel(List<String> agg) throws ArlasExce
         return aggregationModel;
     }
 
+    private static HitsFetcher getHitsFetcher(String fetchHitsString) throws ArlasException {
+        HitsFetcher hitsFetcher = new HitsFetcher();
+        if (StringUtil.isNullOrEmpty(fetchHitsString)) {
+            throw new BadRequestException("fetchHits should not be null nor empty");
+        }
+        Pattern p = Pattern.compile("(\\d*)(\\()(.*)(\\))");
```
laurent-thiebaud-gisaia: If the pattern is static, you shouldn't compile it each time but it should be a static property

arlas-core/src/main/java/io/arlas/server/dao/ElasticCollectionReferenceDaoImpl.java
sfalquier: https://www.dictionary.com/browse/indices

https://www.elastic.co/guide/en/elasticsearch/reference/current/indices.html

I think we have used `indeces` instead of `indices` many times in ARLAS-server and we may correct this in a dedicated PR. Since we have never used `indexes`, we should stay with `indices`.

arlas-core/src/main/java/io/arlas/server/core/FluidSearch.java
MohamedHamouGisaia: In addition to feeding the list, there is also an action of sorting on a field which can't be in a map

arlas-core/src/main/java/io/arlas/server/utils/ParamsParser.java
MohamedHamouGisaia: Absolutely

arlas-core/src/main/java/io/arlas/server/dao/ElasticCollectionReferenceDaoImpl.java
sfalquier: Still `indeces` instead of `indices`?

PR #406, opened 2019-02-27
Refactor Size and Sort

### LINKED ISSUE #405: Refactor 'Size' and 'Sort' classes
The aim of this issue is to combine The classes `Sort` and `Size` in one single class `Page`
From : 
```
public class Size {
     public Integer size;
     public Integer from;
}

public class Sort {
     public String sort;
     public String searchAfter;
}
```
to 

```
public class Page {
     public Integer size;
     public Integer from;
     public String after;
     public String sort;
}
```

MohamedHamouGisaia:
fix #405


arlas-core/src/main/java/io/arlas/server/app/Documentation.java
```diff
@@ -30,17 +30,44 @@
...
-            "- For a geodistance sort, specify the point, from which the distances are calculated, as follow : 'geodistance:lat lon'" +
+            "> **Note 2**: The order of fields matters." +
+            "\n \n" +
+            "> **Note 3** ***geodistance sort***: Sorts the hits centroids by distance to the given **{lat} {lon}** (ascending distance sort). It can be specified at most 1 time." +
+            "\n \n" +
+            "> __**Example 1**__: sort=`age,-timestamp`. Resulted hits are sorted by age. For same age hits, they are decreasingly sorted in time." +
+            "\n \n" +
+            "> __**Example 2**__: sort=`age,geodistance:89 179`. Resulted hits are sorted by age. For same age hits, they are sorted by closest distance to the point(89°,179°)" +
+            "\n \n";
+    public static final String PAGE_PARAM_AFTER = "List of values of fields present in sort param that are used to search after. " +
```
laurent-thiebaud-gisaia: Shouldn't it be put in a property file?

arlas-core/src/main/java/io/arlas/server/model/request/Page.java
```diff
@@ -19,7 +19,9 @@
 
 package io.arlas.server.model.request;
 
-public class Size {
+public class Page {
     public Integer size;
     public Integer from;
+    public String sort;
```
laurent-thiebaud-gisaia: If you only sort the result, it is strange to use a Page whereas you want all results, isn't it?

arlas-core/src/main/java/io/arlas/server/utils/CheckParams.java
```diff
@@ -257,28 +257,58 @@ public static void checkAggregationIntervalParameter(Aggregation aggregationMode
...
+            /** check compatibility between searchafter with from*/
+            if (page.from != null && page.from != 0) {
+                throw new BadRequestException("after cannot be used if from is higher than 0. If you want to use after, please set from to 0 or keep it empty");
+            }
+            /** check compatibility between searchafter and sort */
+            List<String> searchAfterList = Arrays.asList(page.after.split(","));
+            int searchAfterSize = searchAfterList.size();
+            if (page.sort == null) {
+                throw new BadRequestException("search-after param can not be used whitout sort param");
+            } else {
```
laurent-thiebaud-gisaia: You don't need the else (it adds a useless level)

arlas-core/src/main/java/io/arlas/server/utils/CheckParams.java
```diff
@@ -257,28 +257,58 @@ public static void checkAggregationIntervalParameter(Aggregation aggregationMode
...
+                /** check if sort contains geodistance */
+                if (page.sort.toLowerCase().contains(FluidSearch.GEO_DISTANCE)) {
+                    throw new BadRequestException("searchafter cannot be used when geosorting");
+                }
+                String[] sortList = page.sort.split(",");
+                int sortSize = sortList.length;
+                if (searchAfterSize!=sortSize){
```
laurent-thiebaud-gisaia: missing spaces around !=

arlas-core/src/main/java/io/arlas/server/app/Documentation.java
sfalquier: If we decide to output this content in a property file, it may be done in another PR. By the way, since it's not something that can be updated at runtime, it can stay in source files for the moment.

arlas-core/src/main/java/io/arlas/server/model/request/Page.java
sfalquier: There is a default `size` so you cannot have all sorted results without querying with an appropriate page size for the list you want to get. By the way, it has already been discussed in API meeting, no need to go back to discussion for the moment ;)

PR #415, opened 2019-03-06
Throw appropriate exception when applying filters on stored but non-indexed field

### LINKED ISSUE #357: Throw appropriate exception when applying filters on non-indexed fields
When we query a field that is stored in ES but not indexed, we get an not handled ES exception 

image

MohamedHamouGisaia:
Fix #357


arlas-core/src/main/java/io/arlas/server/core/FluidSearch.java
```diff
@@ -205,6 +205,10 @@ private BoolQueryBuilder filter(Expression expression, String dateFormat) throws
             throw new InvalidParameterException(INVALID_PARAMETER_F);
         }
         String field = expression.field;
+        FieldMD fieldMD = getFieldStatus(field);
```
laurent-thiebaud-gisaia: I don't really understand what is the difference between fieldMD and fieldStatus, why is not always the same?

MohamedHamouGisaia: I first called it FieldStatus before renaming it to FieldMD. I forgot to rename it everywhere

sfalquier:
Still working on it? Maybe it could be rebased/merged or closed?

sfalquier:
@MohamedHamouGisaia Should we close this PR?

PR #462, opened 2019-06-26
Fix updateResponse when no propagation

### LINKED ISSUE #461: Tagger with no propagation does not return status information

alainbodiguel:
Fix #461 #470


scripts/start-tagger.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
sylvaingaudan: If the file is a docker resource, then it should be placed in docker-resources

alainbodiguel: My bad, it's already there. I'll remove it.

MohamedHamouGisaia:
LGTM

PR #489, opened 2019-07-22
Move Utility test classes to src/main

MohamedHamouGisaia:
The aim of this `mv` is to make the test classes (`CollectionTool.java`, `DatasetTool`, ...) available in `arlas-tests.jar`. Some of this classes will be used in arlas-tagger


MohamedHamouGisaia:
@sfalquier I completly agree with you, I'll remove this PR

PR #499, opened 2019-08-21
Authentication mode support 

### LINKED ISSUE #494: [auth-1] Authentication mode support
ARLAS server must support an "authentication mode". It can be enabled or disabled thanks to a configuration key in its configuration.yaml file.
When disabled, no control is done on any request.
When enabled, ARLAS server must support the following additional configuration key:
- list of public resources (i.e. not requiring any auth check)

When enabled, ARLAS server must first check if the requested resource is in the public resources list. If not, it must look for a HTTP header "Authorization: Bearer `JWT token`". If not present, a 401 response code must be sent back to the client.

alainbodiguel:
fix #494 : first part of the auth module


arlas-server/src/main/java/io/arlas/server/app/ArlasServer.java
```diff
@@ -171,6 +172,11 @@ public void run(ArlasServerConfiguration configuration, Environment environment)
             LOGGER.info("Explore API disabled");
         }
 
+        // Auth
+        if (configuration.arlasAuthConfiguration.enabled) {
```
sylvaingaudan: what if `configuration.arlasAuthConfiguration` is null?

arlas-server/src/main/java/io/arlas/server/auth/AuthenticationFilter.java
```diff
@@ -0,0 +1,59 @@
...
+    private final Logger LOGGER = LoggerFactory.getLogger(AuthenticationFilter.class);
+    private final ArlasAuthConfiguration authConf;
+
+    public AuthenticationFilter(ArlasAuthConfiguration conf) {
+        this.authConf = conf;
+    }
+
+    @Override
+    public void filter(ContainerRequestContext requestContext) throws IOException {
+        if (!authConf.publicUris.contains(requestContext.getUriInfo().getPath())) {
```
sylvaingaudan: `authConf.publicUris` is a list and not a hashet, so `contains` is slow.

alainbodiguel: Ok, I'm replacing it with a HashSet

arlas-server/src/main/java/io/arlas/server/app/ArlasServer.java
alainbodiguel: Good point. I'll initialize it with enabled=false if null.

PR #501, opened 2019-08-28
Support geometries indexed as WKT

### LINKED ISSUE #482: Geometries that are indexed with WKT format are systematically considered as geo_points

MohamedHamouGisaia:
Fix #482


arlas-core/src/main/java/io/arlas/server/utils/GeoTypeMapper.java
```diff
@@ -22,70 +22,163 @@
...
+        Envelope affectedBounds = new Envelope(-360, 360, -180, 180);
+        WKTReader wkt = new WKTReader(geometryFactory);
+        Geometry geom = null;
+        try {
+            geom = wkt.read(geometry);
+            List<Coordinate> filteredCoord = Arrays.stream(geom.getCoordinates()).filter(coordinate -> affectedBounds.contains(coordinate)).collect(Collectors.toList());
+            if(filteredCoord.size() != geom.getCoordinates().length){
+                throw new InvalidParameterException(INVALID_WKT_RANGE);
+            }
+            IsValidOp vaildOp = new IsValidOp(geom);
```
alainbodiguel: typo: vailOp -> validOp

scripts/tests-integration-stage.sh
```diff
@@ -181,6 +182,7 @@ export ALIASED_COLLECTION="false"
 if [ "$STAGE" == "REST" ]; then test_rest; fi
 if [ "$STAGE" == "WFS" ]; then test_wfs; fi
 if [ "$STAGE" == "CSW" ]; then test_csw; fi
+if [ "$STAGE" == "REST_WKT_GEOMETRIES" ]; then export WKT_GEOMETRIES="true"; test_rest; fi
```
alainbodiguel: the env variable will keep this value for the following tests, no? Is it not a problem?

MohamedHamouGisaia: True, I didn't see board effect: I wanted to test the WKT_GEOMETRIES=true with the rest stage only.
However it is not a problem for the other test to have WKT_GEOMETRIES=true. 

I'll set it to false for the other stages to keep the precedent behaviour of the other tests

MohamedHamouGisaia: @alainbodiguel it's fixed

docs/arlas-collection-model.md
```diff
@@ -80,6 +80,9 @@ The most important fields are:
     ```
     `labels.status` being the taggable field.
 
+!!! info "Important 3"
+    `geometry_path` field value must have the same format in all documents: It's not recommended to index documents where `geometry_path` value is WKT and others as GeoJson. Otherwise a ParseException will be thrown. It goes the same for `centroid_path`.
```
sfalquier: More than "not recommended", it's "not allowed" to store geometries in `geometry_path` and/or `centroid_path` as WKT and GeoJson at the same time.

arlas-core/src/main/java/io/arlas/server/model/response/Hit.java
```diff
@@ -57,18 +58,19 @@ public Hit(CollectionReference collectionReference, Map<String, Object> source,
         if (collectionReference.params.idPath != null) {
             md.id = "" + MapExplorer.getObjectFromPath(collectionReference.params.idPath, source);
         }
+        CollectionReferenceManager.setCollectionGeometriesType(source, collectionReference);
         if (collectionReference.params.centroidPath != null) {
             try {
                 Object m = MapExplorer.getObjectFromPath(collectionReference.params.centroidPath, source);
-                md.centroid = m != null ? GeoTypeMapper.getGeoJsonObject(m) : null;
+                md.centroid = m != null ? GeoTypeMapper.getGeoJsonObject(m, collectionReference.params.getCentroidType()) : null;
             } catch (ArlasException e) {
```
sylvaingaudan: - e.printStackTrace();
+ LOGGER.error() 
+ a comment explain why this never happens or why we don't have to throw an exception

arlas-core/src/main/java/io/arlas/server/model/response/Hit.java
```diff
@@ -57,18 +58,19 @@ public Hit(CollectionReference collectionReference, Map<String, Object> source,
...
                 e.printStackTrace();
             }
         }
         if (collectionReference.params.geometryPath != null) {
             try {
                 Object m = MapExplorer.getObjectFromPath(collectionReference.params.geometryPath, source);
-                md.geometry = m != null ? GeoTypeMapper.getGeoJsonObject(m) : null;
+                md.geometry = m != null ? GeoTypeMapper.getGeoJsonObject(m, collectionReference.params.getGeometryType()) : null;
             } catch (ArlasException e) {
                 e.printStackTrace();
```
sylvaingaudan: e.printStackTrace();
LOGGER.error()
a comment explain why this never happens or why we don't have to throw an exception

arlas-core/src/main/java/io/arlas/server/utils/GeoTypeMapper.java
```diff
@@ -22,70 +22,167 @@
...
-            //Standard lon/lat array as geo_point field
-            try {
-                geoObject = new Point(((Number) ((ArrayList) elasticsearchGeoField).get(0)).doubleValue(), ((Number) ((ArrayList) elasticsearchGeoField).get(1)).doubleValue());
-            } catch (Exception e) {
-                LOGGER.error("unable to parse geo_point from " + elasticsearchGeoField.getClass() + " :" + elasticsearchGeoField, e);
-                throw new NotImplementedException("Not supported geo_point format found.");
-            }
-        } else if (elasticsearchGeoField instanceof HashMap) {
-            if (((HashMap) elasticsearchGeoField).containsKey("type")) {
+        String parseExceptionMsg = "Unable to parse " + type + " from " + elasticsearchGeoField.getClass() + " :" + elasticsearchGeoField;
```
sylvaingaudan: Unable to parse {value} as valid {type}.

docs/arlas-collection-model.md
sylvaingaudan: in all documents within a same collection

PR #504, opened 2019-09-02
Authorization mode: rules processing

### LINKED ISSUE #495: [auth-2] Authorization mode: rules processing
Depends on #494 

When the "authentication mode" is enabled, ARLAS server must be able to extract and interpret a set of rules from the content of the `JWT token`. A rule specifies a resource path pattern, a verb (GET, PUT...) and a priority (used to decide the rules processing order).
When a rule matches the request, then access to the resource is granted. If no rule matches the request, a 403 response is sent back.

No variable, nor regex injection in this version.

alainbodiguel:
Fix #495

### DIFF
--- a/arlas-server/src/main/java/io/arlas/server/auth/ArlasClaims.java
+++ b/arlas-server/src/main/java/io/arlas/server/auth/ArlasClaims.java
@@ -1,0 +1,61 @@
+/*
+ * Licensed to Gisaïa under one or more contributor
+ * license agreements. See the NOTICE.txt file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Gisaïa licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


arlas-server/src/main/java/io/arlas/server/auth/ArlasClaims.java
```diff
@@ -0,0 +1,61 @@
...
+
+        for (String claim : claims) {
+            String[] splitClaim = claim.split(":");
+            if (splitClaim.length >= 3) {
+                switch (splitClaim[0].toLowerCase().trim()) {
+                    case "rule":
+                        if (splitClaim.length != 4) {
+                            LOGGER.warn("Invalid rule claim format: " + claim);
+                        }
+                        rules.add(new RuleClaim(splitClaim[1], splitClaim[2], Integer.valueOf(splitClaim[3])));
```
MohamedHamouGisaia: Array may be out of index (in case of splitClaim.length == 3)

PR #517, opened 2019-09-13
Publish arlas-admin and arlas-client

### LINKED ISSUE #486: Generate a Java client api and push it in maven repo

### LINKED ISSUE #511: Publish arlas-admin module in cloudsmith

MohamedHamouGisaia:
Fix #511

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -266,6 +266,9 @@ if [ "$SIMULATE" == "NO" ]; then
     # publish arlas-core jar
     cd ${BASEDIR}/arlas-core
     mvn -s ${BASEDIR}/conf/maven/settings.xml deploy
+    # publish arlas-admin jar
+    cd ${BASEDIR}/arlas-admin
+    mvn -s ${BASEDIR}/conf/maven/settings.xml deploy
     cd ${BASEDIR}
 else echo "=> Skip pushing jars in maven repo"; fi
 
--- a/arlas-admin/pom.xml
+++ b/arlas-admin/pom.xml
@@ -12,6 +12,17 @@
     <version>11.6.2-SNAPSHOT</version>
   </parent>
 
+  <distributionManagement>
+    <snapshotRepository>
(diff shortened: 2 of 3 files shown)


conf/maven/arlas-client-deploy-setting.xml
```diff
@@ -0,0 +1,10 @@
+<distributionManagement>
+    <snapshotRepository>
```
laurent-thiebaud-gisaia: Should we publish snapshot to the same repository as release? Do we want to make snapshot public?

release.sh
```diff
@@ -206,6 +206,17 @@ docker run --rm \
...
+mkdir -p target/tmp/java-client
+
+docker run --rm \
+    -e GROUP_ID="$(id -g)" \
+    -e USER_ID="$(id -u)" \
+    --mount dst=/input/api.json,src="$PWD/target/tmp/swagger.json",type=bind,ro \
+    --mount dst=/input/config.json,src="$PWD/conf/swagger/java-config.json",type=bind,ro \
+    --mount dst=/output,src="$PWD/target/tmp/java-client",type=bind \
+	gisaia/swagger-codegen-2.3.1 \
+        -l java --type-mappings GeoJsonObject=Object
```
laurent-thiebaud-gisaia: Can we put the type-mappings into the java-config.json (in order to centralize the configuration)? I mean with something like `typeMappings: ["GeoJsonObject=Object"]`

MohamedHamouGisaia: That would have been better but it is not supported in the list of properties that we can declare in java-config.json

conf/maven/arlas-client-deploy-setting.xml
MohamedHamouGisaia: The release process does not publish the snapshot version
The url of snapshot is the public package too and it's specified, just in case

PR #528, opened 2019-10-03
return one or multiple geometries in _geosearch responses

### LINKED ISSUE #466: [Multi-geometry] return one or multiple geometries in _geosearch responses
Add a queryparam `returned_geometries` in explore API endpoints which is a comma separated geometry field_paths.

This parameter indicates which geometries to return : 
- For `_geosearch`
  - each returned hit fom ES need to be split to `N=sizeOf(returned_geometries)` features in the GeoJson :  100 hits from ES <==>  `N * 100` Features in `_geosearch` response.
 - each feature should contain an attribute `properties.geometry: "path_to_geo_field"` : 
  ```json
   {
      "type": "FeatureCollection",
      "features": [
         {
           "type": "Feature",
           "properties": {
              "feature_type": "hit",
              "geometry_path": "path_to_geometry_field",
              "data": {}
         },
         "geometry": {}
        }
      ]
   }
  ```

- For `_search`
  - `hit.md` should contain only the geometries asked in  `returned_geometries`
- `geometry` & `centroid` are reserved words and point respectively to **geometry_path** and **centroid_path** declared in the collection reference.
- if not specified, only  **geometry_path** is returned. If **geometry_path** is null, then **centroid_path** is returned (this is the current behaviour)
- throw an exception if conflict between  `returned_geometries` and `exclude` parameter

alainbodiguel:
Fix #466


arlas-core/src/main/java/io/arlas/server/model/CollectionReferenceParameters.java
```diff
@@ -89,27 +91,34 @@
...
-    private transient GeoTypeEnum geometryType;
-
-    @JsonIgnore
-    private transient GeoTypeEnum centroidType;
+    private Map<String, GeoTypeEnum> geoTypes = new ConcurrentHashMap<>();
 
     public CollectionReferenceParameters() {
     }
 
+    public void setGeoType(String path, GeoTypeEnum type) {
```
MohamedHamouGisaia: is it possible to keep only `setGeoType` and `getGeoType` methods ? `setGeometryType` can be misleading as a method name : it doesn't say necessarily we set the geometry_path type.

Also if it is possible, maybe we can rename `setGeoType` to `setGeometryType`

alainbodiguel: Done!

alainbodiguel:
Cancelled

alainbodiguel:
Cancelled and replaced with #536

PR #536, opened 2019-10-21
return one or multiple geometries in _geosearch responses

### LINKED ISSUE #466: [Multi-geometry] return one or multiple geometries in _geosearch responses

alainbodiguel:
Fix #466 return one or multiple geometries in _geosearch responses

### DIFF
--- a/arlas-rest/src/main/java/io/arlas/server/rest/explore/search/GeoSearchRESTService.java
+++ b/arlas-rest/src/main/java/io/arlas/server/rest/explore/search/GeoSearchRESTService.java
@@ -21,6 +21,7 @@ package io.arlas.server.rest.explore.search;
 
 
 import com.codahale.metrics.annotation.Timed;
+import io.arlas.server.app.Documentation;
 import io.arlas.server.exceptions.ArlasException;
 import io.arlas.server.model.CollectionReference;
 import io.arlas.server.model.enumerations.OperatorEnum;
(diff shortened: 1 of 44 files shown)


arlas-rest/src/main/java/io/arlas/server/rest/explore/search/GeoSearchRESTService.java
```diff
@@ -131,6 +133,13 @@ public Response geosearch(
                     required = false)
             @QueryParam(value = "exclude") String exclude,
 
+            @ApiParam(name = "returned_geometries",
+                    value = Documentation.PROJECTION_PARAM_RETURNED_GEOMETRIES,
+                    allowMultiple = true,
```
MohamedHamouGisaia: I'll change it to allowMultiple = false

PR #557, opened 2019-11-27
WIP FGA columns filtering

laurent-thiebaud-gisaia:


laurent-thiebaud-gisaia:
This is not the final PR. The work isn't finished yet (some tests to be added) but I think to the filtering is applied where it should be.
This is just to get your feedback, on the design or on things I may have forgotten (this is my first time with arlas-server :) ).
You don't need to read it carrefully, the final merge will be splitted into multiple PRs to be easier to read.

Issue #558, opened 2019-11-28
Fine grain access (column filtering)


laurent-thiebaud-gisaia:
From a header parameter with a list of allowed (coma-separated) fields, filter the rest responses to use and return only these fields, if the header is provided. The collection's id, timestamp, geometry and centroid fields are always allowed.

The following endpoints are impacted in `/explore` :

- [X] /_aggregate
- [X] /_count
- [X] /_describe
- [X] /_list
- [X] /_geoaggregate
- [X] /_geosearch
- [X] /_range
- [X] /_search
- [X] /_tile

If a request contains, in its filter, a forbidden field, then a code 403 should be returned. The following parameters are impacted:
- [X] `f` 
- [X] `q`; moreover if a 'q' parameter contains no target field but the column filter header is present, an exception should be thrown
- [X] `sort`
- [X] `include`
- [X] `agggregations` (main field, fetch_geometry, metrics and fetch_hits).

In search requests, only allowed fields should be returned.

In /explore/_describe & /explore/_list:
- [X] the returned properties should be filtered. If a node has no allowed property, then it is hidden.

In ogc-wfs project:
- [x] `DescribeFeatureType`: the returned properties should be filtered
- [x] `GetFeature`: the returned properties should be filtered. Return 403 if a filter field is forbidden
- [x] `GetPropertyValue`: return a 403 if a requested property isn't allowed. Return 403 if a filter field is forbidden

In opensearch project: 
- [X] only return templates with available fields.

At the end, 
- [x] update the documentation

laurent-thiebaud-gisaia:
FGA should also be applied to v12

PR #559, opened 2019-11-28
WIP Fine Grain Access - Column filtering

laurent-thiebaud-gisaia:
Apply column filtering to explore endpoints
(except "describe").
Update tests accordingly.
See #558


laurent-thiebaud-gisaia:
Logic has been changed -> do not remove forbidden fields, but return 403 if there are forbidden fields.
Creating a new branch accordingly.

PR #562, opened 2019-12-03
Feature/column filter requests

laurent-thiebaud-gisaia:
If a field is not available: return 403.

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/utils/RequestFieldsExtractor.java
+++ b/arlas-core/src/main/java/io/arlas/server/utils/RequestFieldsExtractor.java
@@ -1,0 +1,201 @@
+/*
+ * Licensed to Gisaïa under one or more contributor
+ * license agreements. See the NOTICE.txt file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Gisaïa licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.arlas.server.utils;
+
+import io.arlas.server.exceptions.InternalServerErrorException;
+import io.arlas.server.model.request.*;
+
+import java.util.*;
+import java.util.stream.Stream;
+
+public class RequestFieldsExtractor {
+
+    private static final List<String> GEOSORT_IGNORED_LIST = Arrays.asList("geodistance");
+    public static final String INCLUDE_F = "f";
+    public static final String INCLUDE_Q = "q";
+    public static final String INCLUDE_SEARCH_SORT = "search_sort";
+    public static final String INCLUDE_SEARCH_INCLUDE = "search_include";
+    public static final String INCLUDE_SEARCH_EXCLUDE = "search_exclude";
+    public static final String INCLUDE_AGG_FIELD = "agg_field";
+    public static final String INCLUDE_AGG_FETCH_GEOMETRY = "agg_fetch_geometry";
+    public static final String INCLUDE_AGG_FETCH_HITS = "agg_fetch_hits";
+    public static final String INCLUDE_AGG_METRICS = "agg_metrics";
+    public static final String INCLUDE_RANGE_FIELD = "range_field";
+
+    public static final Set<String> INCLUDE_ALL  = new HashSet(Arrays.asList(INCLUDE_F, INCLUDE_Q, INCLUDE_SEARCH_SORT,INCLUDE_SEARCH_INCLUDE, INCLUDE_SEARCH_EXCLUDE, INCLUDE_AGG_FIELD, INCLUDE_AGG_FETCH_GEOMETRY,
+            INCLUDE_AGG_FETCH_HITS, INCLUDE_AGG_METRICS, INCLUDE_RANGE_FIELD));
+
+    public static Stream<String> extract(Request request, Set<String> includeFields) throws InternalServerErrorException {
+        IRequestFieldsExtractor requestExtractor;
+
+        //comparing classes (and not with "instanceof") to be sure that no subclass with new fields has been added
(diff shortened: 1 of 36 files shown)


arlas-core/src/main/java/io/arlas/server/utils/RequestFieldsExtractor.java
```diff
@@ -0,0 +1,165 @@
...
+        if (Arrays.asList(Request.class, Count.class).contains(request.getClass())) {
+            requestExtractor = new BasicRequestFieldsExtractor();
+        } else if (request.getClass() == Search.class) {
+            requestExtractor = new SearchRequestFieldsExtractor();
+        } else if (request.getClass() == AggregationsRequest.class) {
+            requestExtractor = new AggregationRequestFieldsExtractor();
+        } else if (request.getClass() == RangeRequest.class) {
+            requestExtractor = new RangeRequestFieldsExtractor();
+        } else {
+            throw new InternalServerErrorException("Request type is not supporter");
```
alainbodiguel: "supporter" -> "supported"

laurent-thiebaud-gisaia: Oups thx

arlas-tests/src/test/java/io/arlas/server/rest/tag/TagIT.java
```diff
@@ -249,4 +249,6 @@ public void testUntag() throws InterruptedException {
         LOGGER.error("Assertion failed " + tries + " times, we give up.");
         return false;
     }
+
+    //TODO test with columns filter optionnaly
```
alainbodiguel: This can be skipped as the tagger is removed from future version and not used in this version for the targeted deployment.

laurent-thiebaud-gisaia: With Matthieu we thought we could possibly add some tests if we had time. But if this won't be used, no reason to loose time on it.

arlas-tests/src/test/java/io/arlas/server/rest/explore/AbstractPaginatedTest.java
```diff
@@ -91,14 +90,23 @@ public void testPageFrom() throws Exception {
...
-        handleSortParameter(get("sort", search.page.sort), "Dancer");
+        //an empty column filter is not considered
+        handleSortParameter(post(search, ""), "Dancer");
+        handleSortParameter(get("sort", search.page.sort, ""), "Dancer");
 
         search.page.sort = "geodistance:-50 -110";
-        handleGeoSortParameter(post(search), "-50,-110");
-        handleGeoSortParameter(get("sort", search.page.sort), "-50,-110");
+        handleGeoSortParameter(post(search, "fullname,geodistance"), "-50,-110");
+        handleGeoSortParameter(get("sort", search.page.sort, "fullname,geodistance"), "-50,-110");
```
MohamedHamouGisaia: Same remark above, I think `geodistance` should be systematically added

arlas-rest/src/main/java/io/arlas/server/rest/explore/search/SearchRESTService.java
```diff
@@ -299,6 +308,7 @@ public Response searchPost(
         MixedRequest request = new MixedRequest();
         request.basicRequest = search;
         request.headerRequest = searchHeader;
+        RequestColumnFilter.checkRequest(columnFilter, collectionReference, search, searchHeader);
```
MohamedHamouGisaia: I think we should add the fields in `column-filter` into the `include` parameter

arlas-rest/src/main/java/io/arlas/server/rest/explore/search/GeoSearchRESTService.java
```diff
@@ -487,6 +498,7 @@ public Response geosearchPost(
         request.headerRequest = searchHeader;
```
MohamedHamouGisaia: I think we should add the fields in `column-filter` into the `include` parameter

arlas-core/src/main/java/io/arlas/server/utils/RequestColumnFilter.java
```diff
@@ -0,0 +1,153 @@
...
+            throws InternalServerErrorException, ColumnUnavailableException {
+
+        Set<String> fullColumnFilter = getFullColumnFilter(userColumnFilter, collectionReference);
+        if (RequestColumnFilter.isNoFilterDefined(fullColumnFilter)) {
+            return;
+        }
+
+        checkQHasCol(basicRequest);
+        checkQHasCol(headerRequest);
+
```
MohamedHamouGisaia: `collectionReference.collectionParams` has a `filter` attribute that should be checked it toot

laurent-thiebaud-gisaia: Can you cancel the comment if it is not relevant anymore?

arlas-core/src/main/java/io/arlas/server/utils/RequestFieldsExtractor.java
```diff
@@ -0,0 +1,165 @@
...
+                    )
+                    .flatMap(x -> x);
+        }
+
+        private Stream<String> getSortCols(Search search) {
+            return Optional.ofNullable(search.page).flatMap(p -> Optional.ofNullable(p.sort)
+                    .map(sort -> Arrays.stream(sort.split(",")).map(sortCol -> {
+                        if (sortCol.startsWith("-")) {
+                            return sortCol.substring(1);
+                        } else if (sortCol.contains(":")) {
```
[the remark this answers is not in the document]
laurent-thiebaud-gisaia: After thinking twice to this I should we should rather not consider `geodistance` by extracting the sort fields, because `geodistance` isn't really a field. Are you OK with it?

arlas-rest/src/main/java/io/arlas/server/rest/explore/search/SearchRESTService.java
laurent-thiebaud-gisaia: Indeed, that's done

laurent-thiebaud-gisaia:
I fixed issues like we talked about.
Return 403 if a field is used and not allowed.
If a q search has no specific field, also return 403.
Includes with wildcards are not checked (whether to return a 403 or not). However only allowed fields can be included.
Also escape column filter to not inject regexp.

laurent-thiebaud-gisaia:
I also removed ParamsParser.getValidColumnFilter() whose aim is not clear. Replaced by ColumnFilterUtil.isValidColumnFilterPresent().

PR #563, opened 2019-12-03
Feature/column filter describes

laurent-thiebaud-gisaia:


arlas-tests/src/test/java/io/arlas/server/rest/explore/DescribeCollectionServiceIT.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package io.arlas.server.rest.explore;
+
+import io.restassured.path.json.JsonPath;
+import io.restassured.response.ValidatableResponse;
+
+import java.util.Optional;
+
+public class DescribeCollectionServiceIT extends AbstractDescribeTest {
```
MohamedHamouGisaia: I think that a test with a non-empty colum-filter is missing

arlas-tests/src/test/java/io/arlas/server/rest/explore/DescribeServiceIT.java
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+package io.arlas.server.rest.explore;
+
+import io.restassured.path.json.JsonPath;
+import io.restassured.response.ValidatableResponse;
+
+import java.util.Optional;
+
+public class DescribeServiceIT extends AbstractDescribeTest {
```
MohamedHamouGisaia: I think that a test with a non-empty colum-filter is missing

arlas-tests/src/test/java/io/arlas/server/rest/explore/DescribeCollectionServiceIT.java
laurent-thiebaud-gisaia: Right, added `testDescribeFeatureWithEmptyColumFilter` in the parent test class.

arlas-tests/src/test/java/io/arlas/server/rest/explore/DescribeServiceIT.java
laurent-thiebaud-gisaia: Right, added `testDescribeFeatureWithEmptyColumFilter` in the parent test class.

laurent-thiebaud-gisaia:
This is finally the first FGA PR to be merged. It only filters the `describe collection`endpoints, now it supports regexp i.a. the column filter may contain "*" anywhere. See `AbstractDescribeTest` for examples.

laurent-thiebaud-gisaia:
Question: I have a doubt. We said we should support regexp; currently I implemented a wildcard i.a. we may filter on "params" or "params* or "*params". Is it OK or should be support real regexp like "params.(.*)"?

arlas-core/src/main/java/io/arlas/server/utils/ColumnFilterChecker.java
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Provide checking functions to apply column filtering to a request (it returns true or false).
+ *
+ * The principle is:
+ * - if the filter is empty, do not filter at all;
+ * - if there is a filter, turn it to regexp predicates
+ * - append some mandatory fields to it (from the collection)
+ * - checking a path is then a simple regexp match
+ */
+public class ColumnFilterChecker {
```
MohamedHamouGisaia: I think the `ColumnFilterChecker` sould be static

arlas-core/src/main/java/io/arlas/server/core/ElasticAdmin.java
```diff
@@ -43,13 +43,20 @@
...
     }
 
     @SuppressWarnings({"unchecked", "rawtypes"})
-    public CollectionReferenceDescription describeCollection(CollectionReference collectionReference) throws IOException {
+    public CollectionReferenceDescription describeCollection(CollectionReference collectionReference) {
+        return this.describeCollection(collectionReference, Optional.empty());
+    }
+
+    @SuppressWarnings({"unchecked", "rawtypes"})
+    public CollectionReferenceDescription describeCollection(CollectionReference collectionReference, Optional<Set<String>> columnFilterPredicates) {
```
MohamedHamouGisaia: I think it will be better to pass the 2nd argument with the same type for `describeCollection` and `describeAllCollections`.
I see that you use `Optional<Set<String>> columnFilterPredicates` in this recursive function, so maybe we can add a new method : 
`public CollectionReferenceDescription describeCollection(CollectionReference collectionReference, Optional<String> columnFilter)` that fetches the predicate first and then calls this current method

laurent-thiebaud-gisaia: It is not possible the way you propose it, because if 2 methods have a signature with same objects but different generic class (ex: void toto(Optional<String>) vs titi(Optional<List<String>>) the compiler cannot distinguish them.
Using Optional<String> in both methods to have the same signature.

laurent-thiebaud-gisaia:
_rebasing_

PR #564, opened 2019-12-04
Feature/column filter ogc

laurent-thiebaud-gisaia:


arlas-tests/src/test/java/io/arlas/server/ogc/wfs/WFSServiceIT.java
```diff
@@ -72,6 +73,98 @@ public void testNoHeaderFilter() throws Exception {
...
+                        new ImmutablePair<>("SERVICE", "WFS"),
+                        new ImmutablePair<>("VERSION", "2.0.0"),
+                        new ImmutablePair<>("COUNT", "1000"),
+                        new ImmutablePair<>("REQUEST", "GetFeature")),
+                        new Filter(),
+                        Optional.of("params.job,params.country,params.city")));
+    }
+
+    @Test
+    public void testGetFeatureWithUnavailableColumn() throws Exception {
```
MohamedHamouGisaia: I would say : `WithNotReturnedColumn` instead of `WithUnavailableColumn`
Because at first, I thought it's a case where the column is unavailable and we throw an exception; which is not the case for this test

ogc-wfs/src/main/java/io/arlas/server/ogc/wfs/WFSRESTService.java
```diff
@@ -70,7 +74,7 @@
     protected WFSToolService wfsToolService;
 
     //TODO extends and create a wfsexploreService not inject it
-    public WFSRESTService(WFSHandler wfsHandler) {
+    public WFSRESTService(WFSHandler wfsHandler, Client client) {
```
MohamedHamouGisaia: `Client` is no longer used, it should be removed :)

laurent-thiebaud-gisaia:
_rebasing_

PR #607, opened 2020-02-20
Bump elasticsearch dependency to 7.6.0

sfalquier:
With improvements since v7.4.x on geo indices, elasticsearch returns the same results when querying with geometry with out of bounds geo points (lon > 180 for example) and when point sequence is clockwise or counter-clockwize.

@mbarbet @MohamedHamouGisaia : ensure that there is no regression on bbox selection on ARLAS-wui.


MohamedHamouGisaia:
@sfalquier I added the discussed modifications to maintain the same behaviour in ARLAS-server

arlas-core/src/main/java/io/arlas/server/utils/GeoUtil.java
```diff
@@ -167,23 +168,29 @@ public static void checkWKT(String wktString) throws InvalidParameterException {
...
         double envelopeEast = envelope.getMaxX();
         double envelopeWest = envelope.getMinX();
+        GeometryFactory geometryFactory = new GeometryFactory(new PrecisionModel(), 4326);
+        Geometry middleWest = geometryFactory.toGeometry(new Envelope(-180, 0, -90, 90));
+        Geometry middleEast = geometryFactory.toGeometry(new Envelope(0, 180, -90, 90));
         if (envelopeEast <= 180 && envelopeWest >= -180) {
             /** longitudes between -180 and 180**/
-            geometries.add(polygon);
+            if ((envelopeEast - envelopeWest) > 180) {
+                geometries = new ArrayList<>(Arrays.asList(middleWest.intersection(polygon), middleEast.intersection(polygon))
```
laurent-thiebaud-gisaia: Do you really need the `new ArrayList<>(...)`? (same comment for other envelopeEast / envelopeWest cases)

MohamedHamouGisaia: You're right, no need in this case

MohamedHamouGisaia: done

PR #609, opened 2020-02-25
Replaced remaining ES prepare statements

alainbodiguel:


alainbodiguel:
I totally agree and also raised this idea.

sfalquier:
We all agree with this. Do not hesitate to push some PR to start refactoring the codebase with this aim if you have to modify existing stuff or if you have to implement new one... or if you have some free time (lucky you ;) )

PR #611, opened 2020-02-28
New Rest High Level Client for ES

alainbodiguel:
Next step will to replace TransportClient with the new client

### DIFF
--- a/arlas-core/pom.xml
+++ b/arlas-core/pom.xml
@@ -74,6 +74,18 @@
             <version>${elastic.version}</version>
         </dependency>
 
+        <dependency>
+            <groupId>org.elasticsearch.client</groupId>
+            <artifactId>elasticsearch-rest-high-level-client</artifactId>
+            <version>${elastic.version}</version>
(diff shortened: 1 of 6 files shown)


conf/configuration.yaml
```diff
@@ -73,6 +73,7 @@ arlas-index:  ${ARLAS_ELASTIC_INDEX:-.arlas}
 elastic-enable-ssl: ${ARLAS_ELASTIC_ENABLE_SSL:-false}
 elastic-credentials: ${ARLAS_ELASTIC_CREDENTIALS:-user:password}
 elastic-compress: ${ARLAS_ELASTIC_COMPRESS:-true}
+elastic
```
MohamedHamouGisaia: is something missing Or it's set to undefined to keep using the deprecated config attributes?

alainbodiguel: oops! it's a mistake

MohamedHamouGisaia: You have my approval

PR #612, opened 2020-03-04
Using New Rest High Level Client for ES

alainbodiguel:


arlas-core/src/main/java/io/arlas/server/utils/ElasticClient.java
```diff
@@ -0,0 +1,271 @@
...
+        throw new InternalServerErrorException(e.getMessage());
+    }
+
+    @Override
+    protected void finalize() throws Throwable {
+        super.finalize();
+        this.close();
+    }
+
+    public void close() {
```
laurent-thiebaud-gisaia: I think the class may implement Closable, maybe it can be usefull someday

arlas-core/src/main/java/io/arlas/server/utils/ElasticClient.java
```diff
@@ -0,0 +1,271 @@
...
+                throw new NotFoundException("Index " + index + " does not exist.");
+            }
+        }
+        LOGGER.warn("Exception while communicating with ES: " + e.getMessage(), e);
+        throw new InternalServerErrorException(e.getMessage());
+    }
+
+    @Override
+    protected void finalize() throws Throwable {
+        super.finalize();
```
laurent-thiebaud-gisaia: `Finalize` is not always called, will it be OK for the resources?

conf/configuration.yaml
```diff
@@ -66,13 +66,13 @@ logging:
...
 ########################################################
 # Configuration of the datasource
-elastic-nodes: ${ARLAS_ELASTIC_NODES:-localhost:9300}
-elastic-sniffing: ${ARLAS_ELASTIC_SNIFFING:-false}
-elastic-cluster: ${ARLAS_ELASTIC_CLUSTER:-elasticsearch}
 arlas-index:  ${ARLAS_ELASTIC_INDEX:-.arlas}
-elastic-enable-ssl: ${ARLAS_ELASTIC_ENABLE_SSL:-false}
-elastic-credentials: ${ARLAS_ELASTIC_CREDENTIALS:-user:password}
-elastic-compress: ${ARLAS_ELASTIC_COMPRESS:-true}
+elastic:
```
sfalquier: Changing ARLAS-server configuration may be considered as a breaking change, right? Perhaps, we can change configuration keys in a further PR with some other breaking changes (API for example).

cc @MohamedHamouGisaia

MohamedHamouGisaia: Indeed it will be a breaking change. Especially for the ones that configure ARLAS-server using the yaml file. Howevern, deployments using ENV Variables will not be affected

I agree, maybe we can change the keys in v13.7.0 of ARLAS-server. What do you think @alainbodiguel

alainbodiguel: In the code, I support both the old and new format. Is that be ok ? Else, there is no problem with postponing this change.

arlas-core/src/main/java/io/arlas/server/utils/ElasticClient.java
alainbodiguel: Considering that the REST client is not permanently connected to ES (http requests), it will have no impact on ES. And since the finalize method is supposed to be called when stopping the application, I assume it will not be an issue.

conf/configuration.yaml
MohamedHamouGisaia: My bad, I didn't see that
In this case there is no need to postpone the change :)

docs/arlas-server-configuration.md
```diff
@@ -76,13 +76,11 @@ docker run -ti -d \
...
 | --- | --- | --- | ---     |
-| ARLAS_ELASTIC_NODES       | elastic-nodes       | localhost:9300 | coma separated list of elasticsearch nodes as host:port values |
+| ARLAS_ELASTIC_NODES       | elastic-nodes       | localhost:9200 | coma separated list of elasticsearch nodes as host:port values |
 | ARLAS_ELASTIC_SNIFFING    | elastic-sniffing    | false          | allow elasticsearch to dynamically add new hosts and remove old ones (*)|
-| ARLAS_ELASTIC_CLUSTER     | elastic-cluster     | elasticsearch  | clustername of the elasticsearch cluster that is used for storing ARLAS configuration |
 | ARLAS_ELASTIC_INDEX       | arlas-index         | .arlas         | name of the index that is used for storing ARLAS configuration |
 | ARLAS_ELASTIC_ENABLE_SSL  | elastic-enable-ssl  | false          | use SSL to connect to elasticsearch |
 | ARLAS_ELASTIC_CREDENTIALS | elastic-credentials | user:password  | credentials to connect to elasticsearch |
-| ARLAS_ELASTIC_COMPRESS    | elastic-compress    | true           | transport.compress parameter of the elasticsearch connection |
 | ARLAS_ELASTIC_SKIP_MASTER | elastic-skip-master | true           | Skip dedicated master in Rest client |
```
MohamedHamouGisaia: Maybe we can add a note saying that `ARLAS_ELASTIC_CLUSTER` and `ARLAS_ELASTIC_COMPRESS` are not used anymore

conf/configuration.yaml
sfalquier: 👍

sfalquier: Maybe we could add a note in the documentation to explain that old config keys are deprecated.

MohamedHamouGisaia:
@alainbodiguel, @mbarbet , @sfalquier 
I tested this branch with ARLAS-front for each demo, everything works fine

PR #617, opened 2020-03-19
Trace element name error in the rendered xml and in server logs

### LINKED ISSUE #615: WFS GetFeature and GetFeatureById can return an invalid XML without throwing an error
writing invalid element names such as `@version` for example, is invalid from an XML perspective, but ARLAS-server doesn't detect the error.

We end up having an xml parsing error on browsers without any information of the cause
image

MohamedHamouGisaia:
Fix #615

### DIFF
--- a/ogc-common/src/main/java/io/arlas/server/ogc/common/utils/XmlUtils.java
+++ b/ogc-common/src/main/java/io/arlas/server/ogc/common/utils/XmlUtils.java
@@ -27,7 +27,9 @@ import io.arlas.server.model.response.CollectionReferenceDescriptionProperty;
 import io.arlas.server.model.response.ElasticType;
 import io.arlas.server.utils.FilterMatcherUtil;
 import io.arlas.server.utils.MapExplorer;
+import io.arlas.server.utils.StringUtil;
 import io.arlas.server.utils.TimestampTypeMapper;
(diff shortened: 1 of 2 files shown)


ogc-common/src/main/java/io/arlas/server/ogc/common/utils/XmlUtils.java
```diff
@@ -157,8 +164,22 @@ private static void writeElementForType(XMLStreamWriter writer, String nameToDis
         }
     }
 
-    private static void writeElement(XMLStreamWriter xmlStream, String nameToDisplay, String value, String uri, String prefix) throws XMLStreamException {
+    private static void writeElement(XMLStreamWriter xmlStream, String nameToDisplay, String value, String uri, String prefix) throws XMLStreamException, ArlasException {
         nameToDisplay = replacePointPath(nameToDisplay);
+        if (nameToDisplay != null && nameToDisplay.length() > 0) {
```
laurent-thiebaud-gisaia: If you have apache lib, there is this StringUtils.isEmpty() or isBlank() (also trims) ;-)

MohamedHamouGisaia: Fixed

PR #620, opened 2020-03-26
Removal of ES doc type

alainbodiguel:


MohamedHamouGisaia:
We should remove `type` from documentation in `arlas-collection-model.md` file

PR #621, opened 2020-03-31
Removal of ES dependency in arlas-rest

alainbodiguel:
Moving code to ExploreServices (not writing new) in order to remove dependency to ES in REST services

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/services/ExploreServices.java
+++ b/arlas-core/src/main/java/io/arlas/server/services/ExploreServices.java
@@ -29,19 +29,19 @@ import io.arlas.server.exceptions.BadRequestException;
 import io.arlas.server.exceptions.InvalidParameterException;
 import io.arlas.server.managers.CollectionReferenceManager;
 import io.arlas.server.model.CollectionReference;
+import io.arlas.server.model.Link;
 import io.arlas.server.model.enumerations.CollectionFunction;
 import io.arlas.server.model.enumerations.ComputationEnum;
 import io.arlas.server.model.enumerations.GeoTypeEnum;
 import io.arlas.server.model.enumerations.OperatorEnum;
 import io.arlas.server.model.request.*;
-import io.arlas.server.model.response.AggregationMetric;
-import io.arlas.server.model.response.AggregationResponse;
-import io.arlas.server.model.response.ComputationResponse;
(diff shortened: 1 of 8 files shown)


arlas-core/src/main/java/io/arlas/server/utils/UriInfoUtil.java
```diff
@@ -0,0 +1,79 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package io.arlas.server.utils;
+
+import javax.ws.rs.core.UriInfo;
+
+public class UriInfoUtil {
```
laurent-thiebaud-gisaia: I personally find the name may be misleading, I expect static methods from an Util class, whereas this is a wrapper.

arlas-core/src/main/java/io/arlas/server/services/ExploreServices.java
```diff
@@ -202,8 +214,154 @@ public ComputationResponse compute(MixedRequest request, CollectionReference col
         return computationResponse;
     }
 
+    public Hits getSearchHits(MixedRequest request, CollectionReference collectionReference, Boolean flat, UriInfo uriInfo, String method) throws ArlasException {
```
laurent-thiebaud-gisaia: I suppose you only moved the code and didn't change anything?

alainbodiguel: Indeed, you can compare from the original REST services.

arlas-core/src/main/java/io/arlas/server/utils/UriInfoUtil.java
alainbodiguel: You're right. Would it make more sense to rename the class to UriInfoWrapper or reimplement it with static methods?

laurent-thiebaud-gisaia: I don't know how it is used, I feel like there will be more boilerplate code with static methods, but this is your choice :-)

alainbodiguel: Ok. Renamed to UriInfoWrapper

PR #636, opened 2020-06-24
Collection filter

### LINKED ISSUE #635: Collection filter
We need to ensure that the following endpoints take into account the authorised collections for the authenticated user if relevant:

explore/_list
collections
collections/_export
collections/_import

alainbodiguel:
closes #635

### DIFF
--- a/arlas-admin/src/main/java/io/arlas/server/task/CollectionAutoDiscover.java
+++ b/arlas-admin/src/main/java/io/arlas/server/task/CollectionAutoDiscover.java
@@ -57,7 +57,7 @@ public class CollectionAutoDiscover extends Task implements Runnable {
             List<CollectionReferenceDescription> discoveredCollections = collectionDao.getAllIndicesAsCollections();
             List<CollectionReferenceDescription> existingCollections;
             try {
-                existingCollections = collectionDao.describeAllCollections(collectionDao.getAllCollectionReferences(), Optional.empty());
+                existingCollections = collectionDao.describeAllCollections(collectionDao.getAllCollectionReferences(Optional.empty()), Optional.empty());
             } catch (Exception e) {
                 existingCollections = new ArrayList<>();
             }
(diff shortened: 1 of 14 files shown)


sfalquier:
Since the collectionFilter is a kind of document filter like the partitionFilter, I wonder if these concepts should not be merged and managed in the same way...

alainbodiguel:
To be discussed at an ARLAS explo meeting maybe.

MohamedHamouGisaia:
@alainbodiguel, 
In the `explore` API and the `csw` endpoint, the `column-filter` header can, actually, filter columns.
For instance, if we want only **collection1** & **collection3** to be returned, we can write 
```
Column-Filter : ‘collection1:*,collection3:*’
```
So I think there is no need to add a `collection-filter` to those endpoints.

However, there is no collection filtering in `collection` API, so maybe we can keep it for this one.

alainbodiguel:
@MohamedHamouGisaia 
The collection filter is a prefix (i.e. "col1" gives access to "col1_foo" and "col1_bar"). Does the column filter allow for the same logic?
Moreover, it wouldn't make sens the set both `Column-Filter` and `arlas-collection-filter` to achieve collection filtering. It would be preferable to set only one header to achieve this filtering (whichever it is).

MohamedHamouGisaia:
@alainbodiguel 
no the prefix logic is not implemented in column filter.
Yes I agree that we should set one header to achieve collection filtering

alainbodiguel:
Reusing column filter as discussed

alainbodiguel:
New version, with collection part of the columnFilter supporting wildcard.

MohamedHamouGisaia:
We should add a `column-filter` section in `arlas-api-collection.md` doc file as it's the case for [arlas-api-exploration.md](https://github.com/gisaia/ARLAS-server/blob/d0d3743352dfe7db8df321991e52f539c5855036/docs/arlas-api-exploration.md#column-filtering)

MohamedHamouGisaia:
Otherwise, LGTM

alainbodiguel:
> We should add a `column-filter` section in `arlas-api-collection.md` doc file as it's the case for [arlas-api-exploration.md](https://github.com/gisaia/ARLAS-server/blob/d0d3743352dfe7db8df321991e52f539c5855036/docs/arlas-api-exploration.md#column-filtering)

Done. Can you check please?

PR #640, opened 2020-07-05
Extracting super class for ArlasServerConfiguration

alainbodiguel:
To be used, if possible, in :

- Tagger
- Persistence
- Collection server
- City server
- Subscriptions

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/app/ArlasBaseConfiguration.java
+++ b/arlas-core/src/main/java/io/arlas/server/app/ArlasBaseConfiguration.java
@@ -1,0 +1,65 @@
+/*
+ * Licensed to Gisaïa under one or more contributor
+ * license agreements. See the NOTICE.txt file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Gisaïa licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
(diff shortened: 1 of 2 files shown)


arlas-core/src/main/java/io/arlas/server/app/ArlasBaseConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+package io.arlas.server.app;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.smoketurner.dropwizard.zipkin.ZipkinFactory;
+import io.arlas.server.exceptions.ArlasConfigurationException;
+import io.dropwizard.Configuration;
+import io.federecio.dropwizard.swagger.SwaggerBundleConfiguration;
+
+public class ArlasBaseConfiguration extends Configuration {
+
```
mbarbet, line 29: We could add this part of conf, present in arlas-persistence basecode
```
@NotNull
    @JsonProperty("arlas_cors")
    public ArlasCorsConfiguration arlasCorsConfiguration;
```

alainbodiguel: I'll do that in another PR when I have assessed the impact on the other microservices.

PR #643, opened 2020-07-30
No longer escaping replacement value in Auth header variables.

alainbodiguel:


arlas-core/src/main/java/io/arlas/server/app/ArlasAuthConfiguration.java
```diff
@@ -54,10 +57,20 @@
 
     private String publicRegex;
     public String getPublicRegex() {
+        // [swagger.*:*, persist.*:GET/POST/DELETE}]
+        List<String> pathToVerbs = this.publicUris.stream().map(uri -> {
+            String path = uri.split(":")[0];
+            String verbs = uri.split(":")[1];
```
MohamedHamouGisaia, line 63: Maybe we should check if the uri.split(':') length is >= 2 to avoid an eventuak `IndexOutofbounds` exception
unless we are sure that the uri will always have the 'verb' part

conf/configuration.yaml
```diff
@@ -147,7 +147,7 @@ opensearch:
 ########################################################
 arlas_auth:
   enabled: ${ARLAS_AUTH_ENABLED:-false}
-  public_uris: [${ARLAS_AUTH_PUBLIC_URIS:-swagger.*}]
+  public_uris: [${ARLAS_AUTH_PUBLIC_URIS:-swagger.*:*,swagger.*:GET/DELETE}]
```
alainbodiguel, line 150: is this example ok?

arlas-core/src/main/java/io/arlas/server/app/ArlasAuthConfiguration.java
mbarbet: A check is present in configuration validation

sfalquier:
Still LGTM as far as I understand.

PR #658, opened 2020-09-28
Allow a difference of 3 between '_geoaggregate/{geohash}'  precision …

### LINKED ISSUE #656: allow a difference of 3 between '_geoaggregate/{geohash}' precision and geohash aggregation precision
## Current behaviour

When requesting `_geoaggregate/{geohash}` endpoint, the precision of the `{geohash}` path parameter (**P1**) has an impact on the precision of the geoaggregation we define in `agg` query parameter (**P2**)

It should be *0 <= **P2** - **P1** <= 2*. 

Otherwise, if this difference is greater than or equal 3, the request will generate _32x32x32=32786_ buckets (for a difference of 3) which is greater than 10000 buckets limitations.

32 being the number of tiles of precision N split from a tile of precision N-1

## Proposed feature
In order to allow a precision difference (**D**= **P2** - **P1**) of 3
- We split the `{geohash}` into 4 similar bboxes
- We apply a geoash aggregation with a precision P2 = P1 + 3 on each bbox, returning 32786 / 4 = 8192 buckets maximum
- We should apply the 4 requests parallelly to ES (Multi-threading)
- Join the results of the 4 requests and return one list

alainbodiguel:
… and geohash aggregation precision
Closes #656

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/exceptions/ArlasExceptionMapper.java
+++ b/arlas-core/src/main/java/io/arlas/server/exceptions/ArlasExceptionMapper.java
@@ -25,6 +25,7 @@ import org.slf4j.LoggerFactory;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.ExceptionMapper;
 import javax.ws.rs.ext.Provider;
+import java.util.Arrays;
 
 @Provider
 public class ArlasExceptionMapper implements ExceptionMapper<ArlasException> {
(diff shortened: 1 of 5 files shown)


MohamedHamouGisaia (approved):

sfalquier (approved):
Except the disrespect of the 7 golden rules of git commit (https://chris.beams.io/posts/git-commit/) LGTM as far as I understand.

mbarbet (approved):

PR #675, opened 2020-11-02
Trim geohash top and right borders

### LINKED ISSUE #673: Tiled geoaggregations counts data on the border of geohashes twice

MohamedHamouGisaia:
When there are points indexed on a border of 2 geohashes, let's say S & E, the points  will be returned with the /geohash/s & /geohash/e endpoints

To avoid duplicating this data, the geohash is trimmed with (10^-8) precision. This precision is chosen in accordance with points indexation precision that is around 10^-7

Fix #673


alainbodiguel (approved):

sebbousquet (approved):

sfalquier (approved):
LGTM

PR #694, opened 2021-03-08
GeoTile Aggregation

### LINKED ISSUE #449: _(geo)aggregate :  Add geotile aggregation type
As a user of ARLAS-server `_aggregate` service

I want to be able to aggregate geo-points using a `map tile` grid *({zoom}/{x}/{y})* with a large precision range (larger than the geohash grid aggregation precision range)

in order to smooth the precision changes while zooming in (in a map client for example)

According to the documentation 
https://www.elastic.co/guide/en/elasticsearch/reference/7.11/search-aggregations-bucket-geotilegrid-aggregation.html

Technically this means :
- [x] new aggregation type **geotile**
  - datehistogram
  - histogram
  - term
  - geohash
  - **geotile**

- [x] new endpoint `@Path("{collection}/_geoaggregate/{z}/{x}/{y}")
   - so we need to transform {z}/{x}/{y} EPSG 3857 tiles to a bounding box as top_left and bottom_right. (same behaviour of the endpoint `@Path("{collection}/_geoaggregate/{geohash}")`)
- [x] For this new agg type, we need to pass a precision in the `agg` query param, in the `interval` attribute (default value `z+3` if not specified)
   - this precision corresponds to zoom level of the aggregation, that should be larger than or equal to {z} in the path param
   - the maximum difference between the precision and {z} in the path param is **6**
   - If we want a difference of 7 we need to split the  bbox resulting from the {z}/{x}/{y} tile into 4 similar bboxes the same way we did #656 
- [x] How to handle returned geometries by the ES aggregation ?
  - The response of ES is 
```
   "aggregations": {
    "large-grid": {
      "buckets": [
        {
          "key": "8/131/84",
          "doc_count": 3
        },
        {
          "key": "8/129/88",
          "doc_count": 2
        },
        {
          "key": "8/131/85",
          "doc_count": 1
        }
      ]
    }
    }
```
We need to transform the `"key"` to a geometry. The geometry to return is to be specified in `aggregated_geometries` subparam of the `agg` param.

For now `aggregated_geometries` have four possible values
- geohash
- geohash_center
- centroid
- bbox

- [x] First thing we need to do, is replace 
`geohash` => `tile`
`geohash_center` => `tile_center`
in order to be generic to the `**geotile**` and `**geohash**`

- [x] Update documentation
- [ ] Integration tests

alainbodiguel:
closes #449

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/app/Documentation.java
+++ b/arlas-core/src/main/java/io/arlas/server/app/Documentation.java
@@ -109,7 +109,7 @@ public class Documentation {
             "\n \n" +
             "   - **{type}:{field}** part is mandatory. " +
             "\n \n" +
-            "   - **interval** must be specified only when aggregation type is datehistogram, histogram and geohash." +
+            "   - **interval** must be specified only when aggregation type is datehistogram, histogram, geotile and geohash." +
             "\n \n" +
             "   - **format** is optional for datehistogram, and must not be specified for the other types." +
             "\n \n" +
@@ -121,13 +121,13 @@ public class Documentation {
             "\n \n" +
             "   - (**order**,**on**) couple is optional for all aggregation types." +
             "\n \n" +
-            "   - **size** is optional for term and geohash, and must not be specified for the other types." +
+            "   - **size** is optional for term and geohash/geotile, and must not be specified for the other types." +
             "\n \n" +
             "   - **include** is optional for term, and must not be specified for the other types." +
             "\n \n" +
             "- {type} possible values are : " +
             "\n \n" +
-            "       datehistogram, histogram, geohash and term. " +
+            "       datehistogram, histogram, geohash, geotile and term. " +
             "\n \n" +
             "- {interval} possible values depends on {type}. " +
             "\n \n" +
@@ -136,6 +136,8 @@ public class Documentation {
             "       If {type} = histogram, then {interval} = {size}. " +
             "\n \n" +
             "       If {type} = geohash, then {interval} = {size}. It's an integer between 1 and 12. Lower the length, greater is the surface of aggregation. " +
+            "\n \n" +
+            "       If {type} = geotile, then {interval} = {size}. It's an integer corresponding to zoom level of the aggregation, that should be larger than or equal to {z} in the path param, and no bigger than {z}+6. " +
             "\n \n" +
             "       If {type} = term, then interval-{interval} is not needed. " +
             "\n \n" +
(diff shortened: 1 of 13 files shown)


arlas-core/src/main/java/io/arlas/server/utils/CheckParams.java
```diff
@@ -66,15 +66,17 @@
     private static final String INVALID_ON_VALUE = "Invalid 'on-' value : ";
     private static final String BAD_COLLECT_FCT_COLLECT_FIELD_NUMBERS = "'collect_field' and 'collect_fct' occurrences should be even.";
     private static final String DATE_NOW = "now";
-    private static final String AGGREGATED_GEOMETRY_NOT_SUPPORTED = "'geohash' & 'geohash_center' are only supported for geohash aggregation type.";
+    private static final String AGGREGATED_GEOMETRY_NOT_SUPPORTED = "'tile' & 'tile_center' are only supported for geohash nd geotile aggregation type.";
```
MohamedHamouGisaia, line 69: nd => and

arlas-tests/src/test/java/io/arlas/server/rest/explore/AbstractGeohashTiledTest.java
```diff
@@ -33,30 +31,56 @@
...
         pwithin = "geo_params.centroid:within:-5,0,0,5";
-        handleGeohashTileDisjointFromPwithin(geohashTilePwithinGet("geohash:geo_params.centroid:interval-3", Arrays.asList(pwithin), "yn"));
+        handleGeohashTileDisjointFromPwithin(tilePwithinGet("geohash:geo_params.centroid:interval-3", Arrays.asList(pwithin), "yn"));
+
+    }
+
+    @Test
+    public void testGeotile() throws Exception {
+        //GEOTILE
+        // precision = geohashLength  ==> we should have one feature maximum
```
MohamedHamouGisaia, line 64: precision = {z}

arlas-tests/src/test/java/io/arlas/server/rest/explore/AbstractGeohashTiledTest.java
```diff
@@ -33,30 +31,56 @@
...
+        handleGeotileGreaterThanPrecision(tileGet("geotile:geo_params.centroid:interval-5", "5/4/3"), 1, "5/4/3");
+
+        // precision > geohashLength  ==> we could have more than one feature
```
MohamedHamouGisaia, line 67: precision > {z}

arlas-core/src/main/java/io/arlas/server/utils/CheckParams.java
alainbodiguel: fixed

arlas-core/src/main/java/io/arlas/server/impl/elastic/core/FluidSearch.java
```diff
@@ -664,6 +673,20 @@ private GeoGridAggregationBuilder buildGeohashAggregation(Aggregation aggregatio
         return geoHashAggregationBuilder;
     }
 
+    // construct and returns the geohash aggregationModel builder
```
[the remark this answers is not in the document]
alainbodiguel: fixed

arlas-core/src/main/java/io/arlas/server/app/Documentation.java
```diff
@@ -213,6 +215,7 @@
     public static final String GEOAGGREGATION_OPERATION = "Aggregate the elements in the collection(s) as features, given the filters and the aggregation parameters.";
     public static final String SHAPEAGGREGATION_OPERATION = "Aggregate the elements in the collection(s) as features, given the filters and the aggregation parameters, and returns a shapefile of it.";
     public static final String GEOHASH_GEOAGGREGATION_OPERATION = "Aggregate the elements in the collection(s) and localized in the given geohash as features, given the filters and the aggregation parameters.";
+    public static final String GEOTILE_GEOAGGREGATION_OPERATION = "Aggregate the elements in the collection(s) and localized in the given geohash as features, given the filters and the aggregation parameters.";
```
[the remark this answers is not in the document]
alainbodiguel: fixed

arlas-tests/src/test/java/io/arlas/server/rest/explore/AbstractGeohashTiledTest.java
alainbodiguel: fixed

arlas-tests/src/test/java/io/arlas/server/rest/explore/AbstractGeohashTiledTest.java
alainbodiguel: fixed

PR #711, opened 2021-06-28
H3 aggregation

alainbodiguel:
Adding H3 to collection params before implementing aggregations. Please evaluate if I did not forget something.

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/core/model/CollectionReferenceParameters.java
+++ b/arlas-core/src/main/java/io/arlas/server/core/model/CollectionReferenceParameters.java
@@ -47,6 +47,9 @@ public class CollectionReferenceParameters implements Serializable {
     @JsonProperty(value = "centroid_path", required = true)
     public String centroidPath;
 
+    @JsonProperty(value = "h3_path", required = false)
+    public String h3Path;
+
     @JsonProperty(value = "timestamp_path", required = true)
     public String timestampPath;
 
--- a/arlas-tests/src/test/resources/results/_describe_base_result.json
+++ b/arlas-tests/src/test/resources/results/_describe_base_result.json
@@ -91,6 +91,93 @@
           "type": "GEO_SHAPE",
           "taggable": false,
           "indexed": true
+        },
+        "h3": {
+          "type": "OBJECT",
+          "properties": {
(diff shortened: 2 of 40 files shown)


arlas-core/src/main/java/io/arlas/server/core/model/CollectionReferenceParameters.java
```diff
@@ -47,6 +47,9 @@
     @JsonProperty(value = "centroid_path", required = true)
     public String centroidPath;
 
+    @JsonProperty(value = "h3_path", required = true)
```
sylvaingaudan, line 50: Having h3 is optional for a collection so the field should also be optional isn't it?

arlas-tests/src/test/resources/results/_describe_base_result.json
```diff
@@ -91,6 +91,93 @@
...
+              "indexed": true
+            },
+            "15": {
+              "type": "KEYWORD",
+              "taggable": false,
+              "indexed": true
+            }
+          },
+          "taggable": false,
+          "indexed": true
```
sylvaingaudan, line 180: Is this correct to have H3 indexed and taggable? isn't it an intermediate json node?

alainbodiguel: I tried this mapping:
```json
        "h3": {
          "enabled": false,
          "properties": {
            "0": {
              "type": "keyword",
              "index": true
            },
```
expecting
```
...h3={indexed=false, taggable=false, type=OBJECT, properties={11={indexed=true, taggable=false, type=KEYWORD}...
```
as a result but got 
```
...h3={indexed=false, taggable=false, type=OBJECT, properties={11={indexed=false, taggable=false, type=KEYWORD}...
```

so unless there is another way to skip the parent node, I fear it is not possible.

arlas-core/src/main/java/io/arlas/server/core/model/CollectionReferenceParameters.java
alainbodiguel: ok, done.

mbarbet (approved):

MohamedHamouGisaia (approved):

PR #716, opened 2021-09-01
Fix "Public uri sends 403 if we provide an invalid token"

### LINKED ISSUE #710: Public uri sends 403 if we provide an invalid token
We need to check public URI in authorization module if a token is provided

alainbodiguel:
Fix #710


sebbousquet (approved):

MohamedHamouGisaia (approved):

alainbodiguel:
PR cancelled, will be replaced by another one.

PR #774, opened 2022-03-30
Add column names support for shape exports

sfalquier:


arlas-rest/src/main/java/io/arlas/server/rest/explore/ExploreRESTServices.java
```diff
@@ -116,7 +117,9 @@ public File toShapefile(FeatureCollection geojson) throws ArlasException {
                             // the geometry column name of schema for saving into a shapefile must be "the_geom"
                             builder.add("the_geom", adesc.getType().getBinding(), CRS_WGS84);
                         } else {
-                            builder.add(getShortColumnName(adesc.getLocalName(), colNum.get()),
+                            String columnName = shapeColumnNames!=null && shapeColumnNames.containsKey(adesc.getLocalName()) ?
```
MohamedHamouGisaia, line 120: if the shapeColumnsNames map is not declared, it could be good if we can use the `fields_descriptions` map as a second attempt, and then the `adesc.getLocalName()` otherwise

what do you think ?

MohamedHamouGisaia:
@sfalquier I was thinking we would use one map (the `fields_description` map) for shape export

arlas-rest/src/main/java/io/arlas/server/rest/explore/ExploreRESTServices.java
sfalquier: IMHO, too risky.
Since fields_description won't respect the max column name length (<10) and cutting it to 10 can lead to duplicate or irrelevant column names, I find it safer to keep the current behavior and just replace with custom column names if available.

MohamedHamouGisaia (approved):

PR #799, opened 2022-06-24
Use GeoPoint Builder

mbarbet:


alainbodiguel (approved):

arlas-core/src/main/java/io/arlas/server/core/impl/elastic/services/ElasticExploreService.java
```diff
@@ -531,7 +533,7 @@ private boolean isAggregatedGeometry(String subName, List<AggregatedGeometryEnum
         return false;
     }
 
-    private Point getGeohashCentre(String geohash) {
```
alainbodiguel, line 534: Maybe we can keep it as is and move to GeoUtil as it would be needed in GeoTypeMapper

PR #807, opened 2022-07-01
Minimal scope for ElasticFluidSearch

mbarbet:


arlas-core/src/main/java/io/arlas/server/core/impl/elastic/services/ElasticFluidSearch.java
```diff
@@ -454,54 +453,71 @@ private void geoDistanceSort(String geoSort, SortOrder sortOrder) throws ArlasEx
...
                 break;
         }
-        aggregations.remove(0);
-        if (aggregations.size() == 0) {
-            return aggregation;
-        }
-        counter++;
+        //add sub aggregation
+        for (int i = 1; i < aggregations.size(); i++) {
+            switch (aggregationModel.type) {
```
MohamedHamouGisaia, line 495: I guess it should be aggregations.get(i) instead of aggregationModel no ?

alainbodiguel (approved):

MohamedHamouGisaia (approved):

PR #831, opened 2022-09-20
Add GMM REST API based on geotile aggregation

QuCMGisaia:
Integrate the GMM library into ARLAS-server through the "{collection}/_gmm/{z}/{x}/{y}" API.

Also centralize the code that performs a geoaggregation from a geotile or geohash as it is reused in this implementation, and will probably be if the GMM is available through the geohash format.


arlas-rest/src/main/java/io/arlas/server/rest/explore/aggregate/GeoAggregateRESTService.java
```diff
@@ -754,6 +756,270 @@ private FeatureCollection getFeatureCollection(MixedRequest request, CollectionR
...
+            // --------------------------------------------------------
+            // ----------------------- EXTRA --------------------------
+            // --------------------------------------------------------
+            @ApiParam(value = "max-age-cache")
+            @QueryParam(value = "max-age-cache") Integer maxagecache
+    ) throws NotFoundException, ArlasException {
+        CollectionReference collectionReference = exploreService.getCollectionReferenceService()
+                .getCollectionReference(collection);
+
+        if (collectionReference == null)
```
alainbodiguel, line 863: Would be better to put body of `if` between { } (to avoid mistakes in the future in case we add a line). The comment is valid for all `if` in the code.

arlas-rest/src/main/java/io/arlas/server/rest/explore/aggregate/GeoAggregateRESTService.java
```diff
@@ -754,6 +756,270 @@ private FeatureCollection getFeatureCollection(MixedRequest request, CollectionR
...
+            throw new NotFoundException(collection);
+
+        // If there is no aggregations, then throw an error
+        if (agg == null || agg.size() == 0)
```
alainbodiguel, line 867: Can be replaced with `if (CollectionUtils.isEmpty(agg))` (same for other cases below)

alainbodiguel (approved):

MohamedHamouGisaia (approved):

PR #850, opened 2022-11-21
Fix : link with search after

mbarbet:


arlas-core/src/main/java/io/arlas/server/core/impl/elastic/services/ElasticFluidSearch.java
```diff
@@ -406,8 +406,7 @@ public FluidSearchService filterSize(Integer size, Integer from) {
 
     @Override
     public FluidSearchService searchAfter(Page page, String after) {
-        requestBuilder.searchAfter(Arrays.stream(after.split(",")).map(FieldValue::of).toList());
-        return this;
+        requestBuilder.searchAfter(Arrays.stream(after.split(",")).map(FieldValue::of).toList());        return this;
```
alainbodiguel, line 409: why removing the new line?

alainbodiguel (approved):

PR #865, opened 2023-01-03
feat: organisation filtering

### LINKED ISSUE #864: Filter collections on organisation name
We expect to be able to filter the access to collections by the name of an organisation given as a header `arlas-organization` in requests.

A request comes in to REST services with optional HTTP headers as follows:
arlas-organization: foo (maybe a list: refered below as [ORG]; only if IAM activated)
column-filter: col1:*,col2:* (refered below as CF)

For a request containing a collection name (COL) as PathParam or QueryParam:
if COL is public then OK
else 
  if [ORG] is present, and COL.org is not in [ORG] then NOK
  else if CF is present (even empty), use it (COL must be in CF)
    else OK

alainbodiguel:
Fix #864

### DIFF
--- a/arlas-tests/src/test/java/io/arlas/server/tests/rest/collections/CollectionServiceIT.java
+++ b/arlas-tests/src/test/java/io/arlas/server/tests/rest/collections/CollectionServiceIT.java
@@ -31,6 +31,7 @@ import org.junit.runners.MethodSorters;
 import java.util.*;
 import java.util.stream.Collectors;
 
+import static io.arlas.commons.rest.utils.ServerConstants.ARLAS_ORGANISATION;
 import static io.arlas.commons.rest.utils.ServerConstants.COLUMN_FILTER;
 import static io.restassured.RestAssured.*;
 import static org.hamcrest.Matchers.*;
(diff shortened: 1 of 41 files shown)


arlas-tests/src/test/java/io/arlas/server/tests/rest/collections/CollectionServiceIT.java
```diff
@@ -100,14 +101,18 @@ public void test02GetAllCollections() throws Exception {
...
+        jsonAsMap.put(CollectionReference.ORGANISATIONS, getOrganisationsJsonAsMap(false));
 
         // PUT new collection 1
         given().contentType("application/json").body(jsonAsMap)
                 .when().put(arlasPath + "collections/collection1")
                 .then().statusCode(200);
 
         // GET all collections
-        getAllCollections(COLLECTION_NAME, COLLECTION_NAME_ACTOR, "collection1");
+//        getAllCollections(COLLECTION_NAME, COLLECTION_NAME_ACTOR, "collection1");
```
MohamedHamouGisaia, line 112: could we keep this test with a null organisation ?

arlas-tests/src/test/java/io/arlas/server/tests/rest/collections/CollectionServiceIT.java
```diff
@@ -100,14 +101,18 @@ public void test02GetAllCollections() throws Exception {
...
+//        getAllCollectionsWithOrg(DataSetTool.DATASET_ORG_OWNER, COLLECTION_NAME, COLLECTION_NAME_ACTOR, "collection1");
```
MohamedHamouGisaia, line 113: is there a reason why these tests are commented ?

MohamedHamouGisaia (approved):

arlas-tests/src/test/java/io/arlas/server/tests/rest/collections/CollectionServiceIT.java
alainbodiguel: The line comment was a mistake

arlas-tests/src/test/java/io/arlas/server/tests/rest/collections/CollectionServiceIT.java
alainbodiguel: The line comment was a mistake

mbarbet (approved):

PR #897, opened 2023-06-22
Feat: do not validate ES SSL certificate

alainbodiguel:


arlas-core/src/main/java/io/arlas/server/core/impl/elastic/utils/ElasticClient.java
```diff
@@ -70,26 +82,88 @@ public class ElasticClient {
...
+
     public ElasticClient(ElasticConfiguration configuration) {
         // disable JVM default policies of caching positive hostname resolutions indefinitely
         // because the Elastic load balancer can change IP addresses
         java.security.Security.setProperty("networkaddress.cache.ttl", "60");
         java.security.Security.setProperty("networkaddress.cache.negative.ttl", "0");
 
+        SSLContext context = null;
+        try {
+            context = SSLContext.getInstance("TLS");
```
QuCMGisaia, line 130: What does that TLS designates?

QuCMGisaia (approved):

arlas-core/src/main/java/io/arlas/server/core/impl/elastic/utils/ElasticClient.java
alainbodiguel: Transport Layer Security: more recent version of SSL

MohamedHamouGisaia (approved):

PR #920, opened 2023-10-13
Feat/jobpermissions

alainbodiguel:


mbarbet (approved):

QuCMGisaia (approved):

arlas-commons/src/main/resources/roles.yaml
```diff
@@ -41,8 +41,10 @@ technicalRoles:
     description:
     - "Download data from ARLAS"
     permissions:
-    - r:/aproc/processes/download/.*:POST
-    - r:/arlas/aproc/processes/download/.*:POST
+    - r:/aproc/processes/download/execution:POST
```
sylvaingaudan, line 44: what is the difference between 
    - r:/aproc/processes/download/.*:POST
 and
    - r:/arlas/aproc/processes/download/.*:POST
? Why two routes for the same endpoint?

arlas-commons/src/main/resources/roles.yaml
```diff
@@ -41,8 +41,10 @@ technicalRoles:
...
+    - r:/arlas/aproc/processes/download/execution:POST
```
sylvaingaudan, line 45: same here: why prefix with `arlas` and no prefix?

PR #1008, opened 2025-02-26
Feat: refactor checkIfAllowedForOrganisations

### LINKED ISSUE #1001: Unnecessary warnings while filtering collections against organisation
In:
```java
    public List<CollectionReference> getAllCollectionReferences(Optional<String> columnFilter, Optional<String> organisations) throws ArlasException {
```
there is a check (`checkIfAllowedForOrganisations`) that throws an exception if a collection within the list is not visible by the organisation:
https://github.com/gisaia/ARLAS-server/blob/f3ce733f1e393ca4e1781ba75a1b2ed992870651/arlas-core/src/main/java/io/arlas/server/core/impl/elastic/services/ElasticCollectionReferenceService.java#L108

The exception is catched and a warning is printed:
https://github.com/gisaia/ARLAS-server/blob/f3ce733f1e393ca4e1781ba75a1b2ed992870651/arlas-core/src/main/java/io/arlas/server/core/impl/elastic/services/ElasticCollectionReferenceService.java#L120

In a normal scenario, the server will print a lot of warnings since it is will often find collections that are not visible from the organisation. This could be fixed by changing `checkIfAllowedForOrganisations` so that it returns true or false and not raise an exception.

mbarbet:
Fix #1001

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/core/impl/elastic/services/ElasticCollectionReferenceService.java
+++ b/arlas-core/src/main/java/io/arlas/server/core/impl/elastic/services/ElasticCollectionReferenceService.java
@@ -103,23 +103,19 @@ public class ElasticCollectionReferenceService extends CollectionReferenceServic
                 }
                 hits = client.search(requestBuilder.build(), CollectionReferenceParameters.class).hits().hits();
                 for (Hit<CollectionReferenceParameters> hit : hits) {
-                    try {
                         CollectionReference colRef = new CollectionReference(hit.id(), hit.source());
-                        checkIfAllowedForOrganisations(colRef, organisations);
-                        if (CollectionUtil.isCollectionPublic(colRef)) {
-                            collections.add(colRef);
-                        } else {
-                            for (String c : allowedCollections) {
-                                if (CollectionUtil.matches(c, hit.id())) {
-                                    collections.add(colRef);
-                                    break;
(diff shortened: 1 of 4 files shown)


sebbousquet (approved):

arlas-core/src/main/java/io/arlas/server/core/impl/elastic/services/ElasticCollectionReferenceService.java
```diff
@@ -103,23 +103,19 @@ public List<CollectionReference> getAllCollectionReferences(Optional<String> col
...
+                                collections.add(colRef);
+                            } else {
+                                for (String c : allowedCollections) {
+                                    if (CollectionUtil.matches(c, hit.id())) {
+                                        collections.add(colRef);
+                                        break;
+                                    }
                                 }
                             }
                         }
```
MohamedHamouGisaia, line 118: shouldn't we throw an exception in else block ?

mbarbet: No, that's exactly what we wanted to remove. We just list the collections, no need to throw an exception if a user can't see it, it's just reading

MohamedHamouGisaia: You are right !

MohamedHamouGisaia (approved):

PR #1012, opened 2025-04-22
Doc/api documentation

WilliGautier:
Changes:
- Replace widdershins api documentation by mkdocs redoc plugin representation

Sides Changes:
- Fix space in documentation for bullet point representation


mkDocs.sh
```diff
@@ -12,9 +12,21 @@ trap clean_exit EXIT
 SCRIPT_PATH=`cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd`
 cd ${SCRIPT_PATH}
 
+# Clean and Copy documentation to target/generated-docs
+rm -rf target/generated-docs
+mkdir -p target/generated-docs
+
+echo "=> Get md documentation"
+cp -r docs/docs/* target/generated-docs
```
QuCMGisaia, line 20: is it normal that you used `docs/docs` and not just `docs` ?

release.sh
```diff
@@ -295,7 +288,6 @@ if [ "$SIMULATE" == "NO" ]; then
     echo "=> Commit release version"
     git add openapi/openapi.json
     git add openapi/openapi.yaml
-    git add docs/api
```
QuCMGisaia, line 298: shouldn't it be `git add docs/docs/api`

mkDocs.sh
WilliGautier: Yes, it's because mkdocs expect a `docs` folder containing the .md at the same location of the `mkdocs.yaml`

And we put all this in a `docs` folder at the project root that can contain any resources related to the documentation. We used this structure for all arlas project documentation.

release.sh
WilliGautier: You're totaly right, it has moved and not reported here, thanx

WilliGautier: Actually here the line is removed as useless. The `openapi.json` file is already stored and versionned in `openapi` folder at root. 

The file is copied with the documentation in the `mkDocs.sh` script when generating the full documentation (used by arlas-docs for example)

sylvaingaudan (approved):

PR #1014, opened 2025-05-01
Feat: add test with is stac model true

### LINKED ISSUE #1011: Update STAC endpoint for AIRS compliance
- [x] add the property `is_stac_model` to collection reference, default is false
- [x] if `is_stac_model` is `True` then
   - [x] the content of `hit.data` goes to the item's root
   - [x] compute the bbox of the geometry and add it to the item's root
   - [x] add `stac_extensions`,  `stac_version` and `links` to the item's root
   - [x] add `"type": "Feature"` in the item's root
- [x] the field names are replaced: the double character `__`  becomes `:`

mbarbet:
-Fix #1011

### DIFF
--- a/arlas-tests/src/test/java/io/arlas/server/tests/DataStac.java
+++ b/arlas-tests/src/test/java/io/arlas/server/tests/DataStac.java
@@ -1,0 +1,194 @@
+/*
+ * Licensed to Gisaïa under one or more contributor
+ * license agreements. See the NOTICE.txt file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Gisaïa licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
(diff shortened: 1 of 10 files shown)


arlas-tests/src/test/java/io/arlas/server/tests/DataStac.java
```diff
@@ -0,0 +1,196 @@
...
+        @JsonProperty("endpoint_url")
+        public String endpointUrl;
+    }
+
+    public final static String GEOMETRY_PATH = "geometry";
+    public final static String CENTROID_PATH = "centroid";
+    public final static String TIMESTAMP_PATH = "properties.datetime";
+
+}
+
```
QuCMGisaia, line 192: extra lines to remove

QuCMGisaia (approved):

MohamedHamouGisaia (approved):

PR #1016, opened 2025-05-09
Chore: refactor elastic explore service

mbarbet:


arlas-core/src/main/java/io/arlas/server/core/impl/elastic/core/ElasticDocument.java
```diff
@@ -31,12 +31,12 @@
...
 public class ElasticDocument {
 
-    public ElasticClient client;
+    private ElasticClient client;
 
     public ElasticDocument(ElasticClient client) {
         this.client = client;
     }
-
+
```
MohamedHamouGisaia, line 39: trailing spaces

PR #1017, opened 2025-05-09
Chore: refactor ESFluidSearch

mbarbet:


arlas-core/src/main/java/io/arlas/server/core/impl/elastic/services/ElasticFluidSearch.java
```diff
@@ -971,81 +950,82 @@ private void setGeoMetricAggregationCollectField(Metric metric) throws ArlasExce
...
+            } else {
+                throw new BadRequestException(ON_NOT_SPECIFIED);
+            }
+        }
+    }
 
-                }else if(aggregationBuilder instanceof TermsAggregation.Builder){
-                    aggregationBuilder = ((TermsAggregation.Builder) aggregationBuilder).order(bucketOrder);
+    private static <T> void orderIfSupported(ObjectBuilder<T> aggregationBuilder, NamedValue<SortOrder> bucketOrder) throws NotAllowedException {
+        if (aggregationBuilder instanceof DateHistogramAggregation.Builder builder ) {
```
MohamedHamouGisaia, line 976: is it possible to summerize the if else blocks with || in one if block ?

arlas-core/src/main/java/io/arlas/server/core/impl/elastic/services/ElasticFluidSearch.java
```diff
@@ -622,302 +615,291 @@ private Builder.ContainerBuilder buildDateHistogramAggregation(Aggregation aggre
...
+        }
+        for (Map.Entry<String, Set<String>> entry : rgs.entrySet()) {
+            String sort = entry.getKey();
+            String[] includes = entry.getValue().toArray(String[]::new);
+            TopHitsAggregation.Builder topHitsBuilder = AggregationBuilders.topHits().size(1)
+                    .source(s -> s.filter(f -> f.includes(Arrays.asList(includes))));
+            for (String field : sort.split(",")) {
+                String unsignedField = field.replaceFirst(ORDER_SIGN_REGEX, "");
+                CollectionUtil.checkAliasMappingFields(client.getMappings(collectionReference.params.indexName), unsignedField);
+                SortOrder order = field.startsWith("+") ? SortOrder.Asc : SortOrder.Desc;
```
MohamedHamouGisaia, line 881: I checked the doc for fertch hits and we apply the order only if the field is preceded by + or -, in case we don't add a sign, no order is to be applied

arlas-core/src/main/java/io/arlas/server/core/impl/elastic/services/ElasticFluidSearch.java
mbarbet: No sorry

MohamedHamouGisaia (approved):

PR #1019, opened 2025-06-10
Add new variable that enables or not the organisations-related checks

MohamedHamouGisaia:
Side changes
- fix default value of ARLAS_CACHE_FACTORY_CLASS in documentation

### DIFF
--- a/arlas-core/src/main/java/io/arlas/server/core/services/CollectionReferenceService.java
+++ b/arlas-core/src/main/java/io/arlas/server/core/services/CollectionReferenceService.java
@@ -16,9 +16,26 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-
 package io.arlas.server.core.services;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+import java.util.Stack;
+import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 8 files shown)


sebbousquet (approved):

arlas-core/src/main/java/io/arlas/server/core/services/CollectionReferenceService.java
```diff
@@ -428,52 +444,57 @@ public void checkIfIndexAllowedForOrganisations(CollectionReference collection,
...
-                                               boolean ownerOnly)
+            Optional<String> organisations,
+            boolean ownerOnly)
             throws CollectionUnavailableException {
-        if (organisations.isEmpty()) {
-            // no header, we'll trust the column filter if any
-            LOGGER.debug("No organisation header");
-            return true;
-        }
+        if (configuration.arlasCheckOrganisations) {
```
sylvaingaudan, line 463: why don't you start by:
```
if not arlasCheckOrganisations:
   return true
```

This would simplify the reading, don't you think?

sylvaingaudan: Also, it is important to make sure that if arlasCheckOrganisations is null, then it is like arlasCheckOrganisations=true !!!

MohamedHamouGisaia:
New strategy discussed with @sylvaingaudan and @sebbousquet : we don't ignore a vulnerability unless we're sure we re not impacted. Which means we can merge PRs with unfixed vulnerabilities

Issue #1055, opened 2026-06-10
Implement STAC Filter Extension


sylvaingaudan:
- [x] Currently, the STAC Service considers only the first collection within the collection list. To do: raise an exception if len(collection)<>1
- [x] Implement STAC Filter Extension
   - [x] declare in the /stac/conformance
      - [x] Item Search Filter: https://api.stacspec.org/v1.0.0/item-search#filter
      - [x] Filter: http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/filter
      - [x] Features Filter: http://www.opengis.net/spec/ogcapi-features-3/1.0/conf/features-filter
      - [x] CQL2 Text: http://www.opengis.net/spec/cql2/1.0/conf/cql2-text
      - [x] CQL2 JSON: http://www.opengis.net/spec/cql2/1.0/conf/cql2-json
      - [x] Basic CQL2: http://www.opengis.net/spec/cql2/1.0/conf/basic-cql2
      - [x] Basic Spatial Functions: http://www.opengis.net/spec/cql2/1.0/conf/basic-spatial-functions
      - [x] Basic Spatial Functions with additional Spatial Literals: http://www.opengis.net/spec/cql2/1.0/conf/basic-spatial-functions-plus
   - [x] implement 
      - [x] basic spatial functions (s_intersects), including additional Spatial Literals
   - [x] do not implement 
      - [x] Functions
      - [x] Arithmetic Expressions
      - [x] Array functions
      - [x] Property-property comparisons
      - [x] Accent and case insensitive
      - [x] Temporal functions
      - [x] Spatial functions
      - [x] Advanced Comaprison Operators (Like, Between and In)
   - [ ] 
   - [x] GET and POST on 
      - [x] /search (STAC only)
      - [x] /collections/{cid}/items
   - [x] /collections/{cid}/queryables
      - [x] for the type, provide [item.json](https://schemas.stacspec.org/v1.0.0/item-spec/json-schema/item.json) and atomic json types (string, datetime)
      - [x] for the description, provide the name of the field, if provided
      ~~- [ ] for post, implement cql2-json only (return 400 otherwise)~~
      - [x] for get, implement cql2-json and cql2-text
   - [x] /queryables?collections=[{cid}] (STAC only)
- [x] Document in the API documentation the limitations of our implementation (e.g. one collection must be provided in the search, not less not more)

PR #1060, opened 2026-06-30
Feature/queryable as query param

mbarbet:
This PR introduces queryable-aware filtering for STAC endpoints.

### What changed

- `ArlasFilterUtils.cql2toArlasFilterList(...)` now accepts `allowedQueryables` and propagates it through all filter conversion helpers. Property normalization now rejects unsupported queryables. 
- `QueryablesBuilder` now exposes `getAllowedQueryables(JsonNode)` to extract the set of queryable field names from the generated queryables schema. The builder also includes small cleanup changes in schema generation. 
- `StacRESTService` now passes `allowedQueryables` down into filter conversion, so query validation happens during request processing. 
- `StacCollectionsRESTService` now:
  - builds queryables for a collection,
  - extracts dynamic query parameters from the request,
  - rejects unsupported queryable parameters with a `BadRequestException`,
  - passes the allowed queryables to downstream search and filter handling. 
- `StacSearchRESTService` now retrieves queryables for the target collection and forwards the allowed set when building item search filters. 
- `StacRESTService` queryable link generation was adjusted, and `getStacFeatureCollection(...)` now takes the allowed queryables set as an additional argument. 
- CI integration coverage was updated to run `FilteringConformanceSuiteIT` in the STAC filter stage. 
- `configuration.yaml` now advertises the relevant CQL2 and queryables conformance classes. 

### Behavioural impact

- Queries using non-queryable fields are now rejected earlier and more explicitly. 
- Dynamic query parameters on `/collections/{id}/items` are now validated against the collection queryables instead of being accepted blindly. 
- STAC property names are still normalized to internal fields, but only if they belong to the allowed queryables set. 

### Tests

Implemented the [OGC API - Features Part 3 filtering](https://docs.ogc.org/is/19-079r2/19-079r2.html) conformance tests by aligning them with the spec scenarios A.1.1 through A.4.5. The test suite now discovers queryables, collection metadata, and unfiltered feature sets locally per test class, to avoid shared global context and making the tests more independent and reliable.

### Remaining Work to Fully Address Ticket #1055
- Write the documentation

### DIFF
--- a/arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/commons/CollectionFilteringMetadata.java
+++ b/arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/commons/CollectionFilteringMetadata.java
@@ -1,0 +1,80 @@
+package io.arlas.server.tests.stac.conformance.ogc.commons;
+import java.util.*;
+import java.util.ArrayList;
+import java.util.List;
+
+public class CollectionFilterMetadata {
+
+    private String collectionId;
+    private String itemsPath;
+    private String queryablesUri;
+    private String sampleQueryable;
+    private String spatialQueryable;
+    private boolean additionalProperties = true;
+    private List<Double> wgs84Bbox = new ArrayList<>();
+    private List<String> supportedCrs = new ArrayList<>();
+
+    public String getCollectionId() {
+        return collectionId;
+    }
+
+    public void setCollectionId(String collectionId) {
+        this.collectionId = collectionId;
+    }
+
+    public String getItemsPath() {
+        return itemsPath;
+    }
+
+    public void setItemsPath(String itemsPath) {
+        this.itemsPath = itemsPath;
+    }
+
+    public String getQueryablesUri() {
+        return queryablesUri;
+    }
+
+    public void setQueryablesUri(String queryablesUri) {
+        this.queryablesUri = queryablesUri;
+    }
+
+    public String getSampleQueryable() {
+        return sampleQueryable;
+    }
+
+    public void setSampleQueryable(String sampleQueryable) {
+        this.sampleQueryable = sampleQueryable;
+    }
+
+    public String getSpatialQueryable() {
+        return spatialQueryable;
+    }
+
+    public void setSpatialQueryable(String spatialQueryable) {
+        this.spatialQueryable = spatialQueryable;
+    }
+
+    public boolean isAdditionalProperties() {
+        return additionalProperties;
+    }
+
+    public void setAdditionalProperties(boolean additionalProperties) {
+        this.additionalProperties = additionalProperties;
+    }
+
+    public List<Double> getWgs84Bbox() {
+        return wgs84Bbox;
+    }
+
+    public void setWgs84Bbox(List<Double> wgs84Bbox) {
+        this.wgs84Bbox = wgs84Bbox == null ? new ArrayList<>() : new ArrayList<>(wgs84Bbox);
+    }
+
+    public List<String> getSupportedCrs() {
+        return supportedCrs;
+    }
+
+    public void setSupportedCrs(List<String> supportedCrs) {
+        this.supportedCrs = supportedCrs == null ? new ArrayList<>() : new ArrayList<>(supportedCrs);
+    }
+}
--- a/arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/commons/CollectionFilteringMetadata.java
+++ b/arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/commons/CollectionFilteringMetadata.java
@@ -1,9 +1,28 @@
+/*
+ * Licensed to Gisaïa under one or more contributor
+ * license agreements. See the NOTICE.txt file distributed with
+ * this work for additional information regarding copyright
+ * ownership. Gisaïa licenses this file to you under
+ * the Apache License, Version 2.0 (the "License"); you may
+ * not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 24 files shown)


arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/filter/FilterConformanceIT.java
```diff
@@ -0,0 +1,157 @@
...
+    // ARLAS-Server does not support custom filter-crs param
+    @DisplayName("Scenario: A.3.6 Conformance Test 11 - /conf/filter/filter-crs-param")
+    void filterCrsParam() {
+
+    }
+
+    @Test
+    @Disabled
+    // ARLAS-Server does not support custom functions
+    @DisplayName("Scenario: A.3.7 Conformance Test 12 - /conf/filter/get-functions")
```
sylvaingaudan: [Conformance Class "Filter"](https://docs.ogc.org/is/19-079r2/19-079r2.html#_conformance_class_filter) provides a flag to tell if it supports functions. Therefore, I believe we must test that the flag=false but we do not have to implement that test, even disabled.

arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/filter/FilterConformanceIT.java
```diff
@@ -0,0 +1,157 @@
...
+                    Map.of("filter-lang", "cql2-text", "filter", invalidFilter),
+                    "application/geo+json");
+            assertEquals(400, invalid.statusCode());
+        }
+    }
+
+    @Test
+    @Disabled
+    // ARLAS-Server does not support custom filter-crs param
+    @DisplayName("Scenario: A.3.6 Conformance Test 11 - /conf/filter/filter-crs-param")
```
sylvaingaudan, line 144: I believe the test should be implemented and the api should return an error ([see case C](https://docs.ogc.org/is/19-079r2/19-079r2.html#filter-crs-param)) if the crs is different from CRS84.

sylvaingaudan (approved):

arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/filter/FilterConformanceIT.java
mbarbet: The flag is an input for the test, it's always false for us, so I just remove the test ?

arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/filter/FilterConformanceIT.java
mbarbet: ok I add it

arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/filter/FilterConformanceIT.java
sylvaingaudan: exactly

arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/commons/CollectionFilteringMetadata.java
```diff
@@ -0,0 +1,100 @@
...
+
+    public List<Map<String, Object>> getUnfilteredFeatures() {
+        return unfilteredFeatures;
+    }
+
+    public void setUnfilteredFeatures(List<Map<String, Object>> unfilteredFeatures) {
+        this.unfilteredFeatures = unfilteredFeatures;
+    }
+
+    private List<Map<String, Object>> unfilteredFeatures = new ArrayList<>();
```
MohamedHamouGisaia: can you place this attribute with the attributes above ?

arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/commons/QueryablesDiscoverySupport.java
```diff
@@ -0,0 +1,132 @@
...
+
+                if (sampleQueryable == null && entry.getKey().equals("params.age")) {
+                    sampleQueryable = entry.getKey();
+                }
+                if (spatialQueryable == null  && entry.getKey().equals("geo_params.geometry")) {
+                    spatialQueryable = entry.getKey();
+                }
+            }
+
+            assertNotNull(sampleQueryable, "No sample queryable found for " + collectionId);
```
MohamedHamouGisaia, line 84: need a assertNotNull For spatialQueryable ?

arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/filter/FilterConformanceIT.java
```diff
@@ -0,0 +1,157 @@
...
+    @Test
+    @DisplayName("Scenario: A.3.3 Conformance Test 8 - /conf/filter/filter-lang-default")
+    void filterLangDefault() {
+        for (CollectionFilteringMetadata metadata : metadataList) {
+            String queryable = metadata.getSampleQueryable();
+
+            Response filtered = get(metadata.getItemsPath(),
+                    Map.of("filter", CqlFilterBuilder.validNullCheck(queryable)),
+                    "application/geo+json");
+            // ARLAS SERVER does not support is null filter, the conformance test should check 200
```
MohamedHamouGisaia, line 86: test should check 200 -> 400 no ?

MohamedHamouGisaia (approved):

arlas-tests/src/test/java/io/arlas/server/tests/stac/conformance/ogc/filter/FilterConformanceIT.java
mbarbet: it should be a 200 yes but with ARLAS it's 400 because we dont support IS NULL as filter

QuCMGisaia (approved):

